PR #191, opened 2018-08-03
Add modifycontractpercent rpc

tjchern:
1.add modify contract_percent rpc
2.delete the restrict that one account just can deploy one contract


tjchern:
have uniform function name, add the percent validation

PR #293, opened 2019-07-04
Support AccountPermissionUpdateContract in TransactionUtils#getOwner

gologo13:
per as title.

this fix should be appreciated in parsing transactions which contain  `AccountPermissionUpdateContract`.


Sh11thead:
LGTM

PR #294, opened 2019-07-04
Move demo and test files in production to test project

gologo13:
Indeed demo classes are really useful to understand how this library works, but should not be included in production code.
Otherwise, jar file contains these demo classes and test file, which is obviously not recommended.


kevinLee2015:
Hi,

First thanks for your contribution, but you can not merge it into master branch directly, you need to merge it to develop branch.

PR #306, opened 2019-09-16
Feature/command optimization

4GPTT:


zhangwenhua-tron:
1. Use jline3 to support command recommended and history.
2. Support setting ownerAddress when creating contract.
3. Use println instead of log output info.
4. The result of command is json object.

PR #327, opened 2019-12-27
Feature/sm2sm3 . Integrate SM2

alberto-zhang:


src/main/java/org/tron/common/utils/Utils.java
```diff
@@ -507,5 +508,17 @@ public static JSONObject printTransactionToJSON(Transaction transaction, boolean
     jsonTransaction.put("txID", txID);
     return jsonTransaction;
   }
+
+  public static boolean confirmEncrption() {
+    System.out.println(
+            "Please confirm encryption module,if input y or Y means default Eckey, other means SM2.");
```
andelf: space after comma

alberto-zhang: ok

PR #337, opened 2020-01-14
fix base58 for consistency(java-tron ,tron factory)

alberto-zhang:
fix base58 for consistency(java-tron ,tron factory)


src/main/java/org/tron/common/crypto/SignInterface.java
```diff
@@ -18,7 +18,7 @@
 
   byte[] getNodeId();
 
-  byte[] Base64toBytes(String signature);
+  byte[] Base64toBytes (String signature);
```
andelf: ？

PR #338, opened 2020-01-19
Unified minimum currency unit is sun

houlei1017:


src/main/java/org/tron/common/utils/TransactionUtils.java
```diff
@@ -247,11 +247,14 @@ public static Transaction setExpirationTime(Transaction transaction) {
...
 
-  public static Transaction setPermissionId(Transaction transaction) throws CancelException {
+  public static Transaction setPermissionId(Transaction transaction, String tipString)
+      throws CancelException {
     if (transaction.getSignatureCount() != 0
         || transaction.getRawData().getContract(0).getPermissionId() != 0) {
       return transaction;
     }
+
+    System.out.println(tipString);
```
andelf: is this debug print?

andelf:
Use uppercase `SUN`

PR #340, opened 2020-02-12
set random ovk if send shielded coin from public address

houlei1017:


andelf:
Is `0x0000000000000000000000000000000000000000000000000000000000000000` acceptable?

Getting a random ovk from a new meaningless address seems redundant.

PR #420, opened 2021-03-17
rm parameter from protocol

sean-liu55:


Federico2014 (approved):

joe192:
diff --git a/src/main/protos/core/Tron.proto b/src/main/protos/core/Tron.proto
index 5c8b5b32..ba26829f 100644
--- a/src/main/protos/core/Tron.proto
+++ b/src/main/protos/core/Tron.proto
@@ -148,9 +148,6 @@ message Account {
   Permission owner_permission = 31;
   Permission witness_permission = 32;
   repeated Permission active_permission = 33;
-
-  //vote power,include frozen bandwidth and frozen energy
-  int64 vote_power_413 = 60;
 }

PR #441, opened 2021-07-21
add EnergyUsed field in TransactionExtention

Lredhdx:


src/main/protos/api/api.proto
```diff
@@ -1195,6 +1195,7 @@ message TransactionExtention {
   bytes txid = 2; //transaction id = sha256(transaction.raw_data)
   repeated bytes constant_result = 3;
   Return result = 4;
+  int64 EnergyUsed = 5;
 }
 
 message BlockExtention {
```
thechillday1, line 1201: 🩸

PR #535, opened 2022-12-29
feat(DynamicEnergy): add dynamic energy apis for wallet

daniel-cao-byte:
Add dynamic energy related protocol buffer types and add estimate energy command.


zhangheng1536 (approved):
LGTM

PR #642, opened 2023-05-22
feat(freezeV2): upgrade stake2.0 APIs

lxcmyf:
Add `CancelAllUnfreezeV2` api and modify `DelegateResource` api for wallet

### DIFF
--- a/src/main/java/org/tron/common/utils/Utils.java
+++ b/src/main/java/org/tron/common/utils/Utils.java
@@ -18,6 +18,7 @@
 
 package org.tron.common.utils;
 
+import com.alibaba.fastjson.JSON;
 import com.alibaba.fastjson.JSONArray;
 import com.alibaba.fastjson.JSONObject;
 import com.google.protobuf.Any;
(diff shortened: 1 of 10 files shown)


src/main/java/org/tron/common/utils/Utils.java
```diff
@@ -618,6 +620,12 @@ public static JSONObject printTransactionToJSON(Transaction transaction, boolean
                             JSONObject.parseObject(
                                     JsonFormat.printToString(unDelegateResourceContract, selfType));
                     break;
+                  case CancelAllUnfreezeV2Contract:
+                    CancelAllUnfreezeV2Contract cancelAllUnfreezeV2Contract =
+                        contractParameter.unpack(CancelAllUnfreezeV2Contract.class);
+                    contractJson = JSON.parseObject(
```
chengtx01, line 626: Why not use JSONObject.parseObject like everywhere else.

lxcmyf: ParseObject is a static method of the parent class, and it is recommended to call static methods directly through the parent class.

chengtx01 (approved):

Mohamad5159 (commented):
ترون انتقال دادم ب تراست ولت نرسیده ب کیف میتونی بااین برنامه کمکم کنی

PR #670, opened 2023-09-27
fix(api): fix API doc link jump

lxcmyf:


halibobo1205 (approved):

jwrct (approved):

cbates8979:
my tron wallet has never loaded idk whats going on or why i have to go through all of this when i came in with funds.why wouldnt my wallet just load like any other transfer

PR #749, opened 2024-09-30
Update config.conf

Posty1211:


317787106:
There is no necessary to change the config. Default value is ok.

PR #823, opened 2025-04-22
fix(information): optimize prompt information and user experience

lxcmyf:
optimize prompt information and user experience


xxo1shine (approved):

src/main/java/org/tron/walletcli/WalletApiWrapper.java
```diff
@@ -445,7 +446,9 @@ private void selectWalletFileByList(List<WalletFile> walletFileList) {
         break;
       }
     } else {
-      wallet = new WalletApi(walletFileList.get(0));
+      WalletFile wf = walletFileList.get(0);
```
shiziwen, line 449: what about size==0?

CodeNinjaEvan (approved):

PR #887, opened 2025-11-07
fix(cmd): Fix Freeze/UnFreezeBalance commands with TP

warku123:
Delete all description about TRON_POWER when freeze balance.

In tron resource model, TP can obtained by freeze balance for both energy and bandwidth. So there are not a way to get TP only.

截屏2025-11-07 16 25 30

And java-tron backend can't receive freeze/unfreeze request with TRON_POWER.

截屏2025-11-07 11 39 28

截屏2025-11-07 11 39 48

So I modify the README and Client.java to fix


lxcmyf:
@warku123 
The tron_power enumeration resource type does exist in the proto file. Because of historical reasons, some tips based on it have not been implemented, but it does not affect the normal use of wallet cli at present. It is recommended to maintain the status quo.

PR #926, opened 2026-04-28
feat(cli): allow known globals after command; query output + contract…

gummy789j:
1. GlobalOptions parses the full argv: after the command token, recognized globals (--network, --output, --wallet, --grpc-endpoint, etc.) are applied and removed from command args; --help/-h after the command are left for the command parser; unknown trailing flags pass through unchanged. Malformed or invalid values for known globals still fail at the global layer (validated globally).

2. Client.requestsJsonOutput no longer stops scanning at the first non-flag so post-command --output json is respected.

3. OutputFormatter adds queryResult(): text mode renders small flat primitive maps as Metadata, larger/nested/array payloads as indented Result: JSON; success/formatMessage-only paths wrap plain strings for TEXT JSON payloads.

4. QueryCommands and ContractCommands use queryResult for RPC-style successes; several protobuf paths now emit not_found when the RPC returns null.

5. Updates standard-cli-contract-spec and qa/contracts.tsv (post-command network succeeds; unknown post-command option stays command-local usage).

6. Adds GlobalOptions, OutputFormatter, and ClientMain tests for the above.

### DIFF
--- a/src/main/java/org/tron/walletcli/cli/GlobalOptions.java
+++ b/src/main/java/org/tron/walletcli/cli/GlobalOptions.java
@@ -1,6 +1,7 @@
 package org.tron.walletcli.cli;
 
-import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 
 public class GlobalOptions {
@@ -41,26 +42,36 @@ public class GlobalOptions {
         boolean networkSeen = false;
         boolean walletSeen = false;
         boolean grpcEndpointSeen = false;
+        List<String> commandArgs = new ArrayList<String>();
         for (int i = 0; i < args.length; i++) {
             String token = args[i];
+            boolean commandSeen = opts.command != null;
 
-            if (!token.startsWith("-")) {
+            if (!commandSeen && !token.startsWith("-")) {
                 opts.command = token.toLowerCase(Locale.ROOT);
-                opts.commandArgs = Arrays.copyOfRange(args, i + 1, args.length);
-                return opts;
-            }
-
-            if ("-h".equals(token)) {
-                opts.help = true;
                 continue;
             }
 
-            if (token.startsWith("-h=")) {
+            if ("-h".equals(token)) {
+                if (commandSeen) {
+                    commandArgs.add(token);
+                } else {
+                    opts.help = true;
+                }
+                continue;
+            }
+
+            if (!commandSeen && token.startsWith("-h=")) {
                 throw new CliUsageException("Option -h does not take a value");
             }
 
             if (!token.startsWith("--")) {
-                throw new CliUsageException("Unknown global option: " + token);
+                if (commandSeen) {
+                    commandArgs.add(token);
+                } else {
+                    throw new CliUsageException("Unknown global option: " + token);
+                }
+                continue;
             }
 
             ParsedLongOption parsed = parseLongOptionToken(token);
@@ -70,6 +81,10 @@ public class GlobalOptions {
                     opts.interactive = true;
                     break;
                 case "help":
+                    if (commandSeen) {
+                        commandArgs.add(token);
+                        break;
+                    }
                     ensureNoInlineValue(parsed, "--help");
                     opts.help = true;
                     break;
@@ -124,9 +139,14 @@ public class GlobalOptions {
                     }
                     break;
                 default:
-                    throw new CliUsageException("Unknown global option: --" + parsed.name);
+                    if (commandSeen) {
+                        commandArgs.add(token);
+                    } else {
+                        throw new CliUsageException("Unknown global option: --" + parsed.name);
+                    }
             }
         }
+        opts.commandArgs = commandArgs.toArray(new String[0]);
         return opts;
     }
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/tron/walletcli/cli/OutputFormatter.java
```diff
@@ -234,6 +225,87 @@ public void keyValue(String key, Object value) {
         recordSuccess(key + " = " + value, data);
     }
 
+    private void renderQueryResult(Object jsonData) {
+        if (shouldRenderAsMetadata(jsonData)) {
+            renderMetadata(jsonData);
+            return;
```
3for, line 231: `queryResult()` in TEXT mode drops both empty objects and message-only results.  
`normalizeJsonData()` wraps non-JSON strings into `{ message: ... }`, and empty protobuf/list payloads may become `{}`.  
Then `shouldRenderAsMetadata()` returns `true`, but `renderMetadata()` (line 242) skips the `message` field, leaving nothing to print—so the output degrades to just `“X successful !!”`.  

This affects the commands switched to `queryResult()` in this commit (e.g., `QueryCommands.java:746`).  
The previous `printMessage` at least printed the raw payload.

**Suggestion:**  
- Make `shouldRenderAsMetadata()` require **at least one non-`message` field**.  
- Otherwise, fall back to `Result:` output, or directly print `message` for message-only cases.

**Suggested tests:**

| Input | After normalize | shouldRenderAsMetadata | renderMetadata output |
|------|----------------|------------------------|-----------------------|
| `"plain text"` (non-JSON) | `{"message":"plain text"}` (Map) | true (only `message`, skipped) | empty → nothing printed |
| `"{}"` (empty JSON, e.g. empty proto) | empty JsonObject | true (size ≤ 12, loop not entered) | empty → nothing printed |
| `{"message":"only"}` | single-key JsonObject | true (only key skipped) | empty → nothing printed |

src/main/java/org/tron/walletcli/cli/GlobalOptions.java
3for, line 102: **Behavior change:**  
`--version` and `--interactive` are now accepted as **post-command global options**, and are handled early in `Client.java:4721` before command execution.  

This means `wallet-cli get-balance --version` returns the version instead of a command usage error.  

The docs seem to allow “known global options can appear after the command”, but `--help` is explicitly reserved for command-level handling.  

**Suggestion:**  
Confirm whether `--version` / `--interactive` should also be restricted to **pre-command only**, for consistency with `--help`.

src/main/java/org/tron/walletcli/cli/OutputFormatter.java
```diff
@@ -234,6 +225,87 @@ public void keyValue(String key, Object value) {
...
+        }
+        out.println();
+        out.println("  Result:");
+        out.println(indentResult(gson.toJson(jsonData)));
+    }
+
+    private String indentResult(String text) {
+        return "  " + text.replace(System.lineSeparator(), System.lineSeparator() + "  ");
```
3for: **Cross-platform bug:**  
`OutputFormatter.indentResult()` uses `System.lineSeparator()`, but Gson always emits `\n` (see `OutputFormatter.java:238-240`):

```java
private String indentResult(String text) {
    return "  " + text.replace(System.lineSeparator(), System.lineSeparator() + "  ");
}
````

`gson.toJson(...)` internally hardcodes `\n` as the line break, regardless of platform.
On Windows, `System.lineSeparator() == "\r\n"`, so `replace()` never matches → multiline JSON is not indented except for the first line.

**Suggestion:**

```java
return "  " + text.replace("\n", "\n  ");
```

(`println(...)` will append the platform-specific newline anyway; using `\n` here for internal formatting is sufficient.)

gummy789j: Fixed. `indentResult()` now indents Gson pretty-printed output by replacing literal `\n`, which matches Gson's internal line separator regardless of the platform line separator.

Added TEXT mode coverage that asserts pretty-printed JSON lines are indented after line feeds.

src/main/java/org/tron/walletcli/cli/GlobalOptions.java
gummy789j: Changed this contract: `--version` and `--interactive` are now treated as top-level mode selectors and are recognized only before the command token. When they appear after a command, they are passed through as command-local args and produce command usage errors such as `Unknown option: --version`.

Updated `standard-cli-contract-spec.md`, added parser/runner tests, and added QA contract cases for post-command `--version` / `--interactive`.

src/main/java/org/tron/walletcli/cli/OutputFormatter.java
gummy789j: Fixed. `shouldRenderAsMetadata()` now requires at least one printable non-`message` field before using the Metadata rendering path, so message-only payloads and empty objects fall back to `Result:` instead of being dropped.

Added regression coverage for non-JSON text (`data.message`) and empty JSON object query results in TEXT mode.

Will-Guan (approved):

zerodevblock-cyber (approved):

Blessedwife6 (approved):

Supermario1975-code (approved):

PR #935, opened 2026-06-16
fix(permission): exclude disabled ContractType 51 and surface unknown ops in  UI

### LINKED ISSUE #934: UpdateAccountPermission: default active permission grants disabled ContractType 51
## Description

In the interactive `updateaccountpermission` flow, when an account has **no existing active permission**, wallet-cli applies a hardcoded default active-permission operations bitmap. That bitmap grants **ContractType 51 (ShieldedTransferContract)**. Since shielded transfer is disabled by default on mainnet and most networks, the node rejects the whole transaction during validation with:

```
CONTRACT_VALIDATE_ERROR, Contract validate error : 51 isn't a validate ContractType
UpdateAccountPermission  failed !!!
```

## Steps to Reproduce

1. Run interactive `updateaccountpermission` on an account that has **no active permission**.
2. Edit only the owner_permission (e.g. add an authorized key); leave active untouched.
3. Preview, confirm, sign, and broadcast.
4. The node returns `51 isn't a validate ContractType` and the transaction fails.

## Root Cause

The hardcoded default operations bitmap in `src/main/java/org/tron/core/manager/UpdateAccountPermissionInteractive.java:108`:

```java
active.setOperations("7fff1fc0033efb0f000000000000000000000000000000000000000000000000");
```

Decoding this bitmap, the granted operations include bit 51:

```
[0,1,2,3,4,5,6,8,9,10,11,12,13,14,15,16,17,18,19,20,30,31,32,33,
 41,42,43,44,45,48,49, 51, 52,53,54,55,56,57,58,59]
                       ^^ ContractType 51 = ShieldedTransferContract
```

Byte index 6 is `0xfb` (binary `1111 1011`), whose bit 3 maps to operation 51. Shielded transfer is disabled by default, so when the node iterates the operations bitmap and hits the unavailable ContractType 51, it rejects the transaction.

## Why the bit is invisible in the UI

`operationsMap` (the human-readable label table at lines 50–88 of the same file) jumps from `49` straight to `52` — there is **no label for 51**. Both the preview and the add/delete menus filter entries with `operationsMap.get(...) != null`. As a result:

- Bit 51 is actually present in the bitmap sent to the node,
- but it is silently hidden from the preview and the edit menu,
- so the user can neither see it nor remove it through the interactive UI.

This is the most subtle part of the bug and makes it very hard to self-diagnose.

## Suggested Fix

1. Correct the default bitmap to clear bit 51: byte index 6 `0xfb` -> `0xf3`, i.e.
   `7fff1fc0033efb0f...` -> `7fff1fc0033ef30f...`
   (keeps 48, 49, 52, 53, 54, 55; only removes 51).
2. (Optional but recommended) Validate/sanitize every set bit in the default bitmap so this class of "present in the bitmap but missing from `operationsMap`, hence invisible and undeletable in the UI" hidden bit cannot recur; or explicitly surface such unknown operations in the UI instead of silently filtering them out.

## Environment

- Affected command: interactive `updateaccountpermission`
- Affected file: `src/main/java/org/tron/core/manager/UpdateAccountPermissionInteractive.java:108`

Federico2014:
## Summary

Fixes two related problems in the interactive `updateaccountpermission` flow, both stemming from how the active-permission operations bitmap is handled.

## 1. Default active permission grants a disabled ContractType 

When an account has **no existing active permission**, the flow applied a hardcoded default operations bitmap that granted **ContractType 51 (ShieldedTransferContract)**. Shielded transfer is disabled on mainnet and most networks, so the node rejected the whole `AccountPermissionUpdate`:

```
CONTRACT_VALIDATE_ERROR, Contract validate error : 51 isn't a validate ContractType
UpdateAccountPermission  failed !!!
```

This triggered even when the user only edited owner_permission, because the default active permission is auto-applied.

**Fix:** clear bit 51 in the default bitmap (`UpdateAccountPermissionInteractive.java`): byte index 6 `0xfb` -> `0xf3`, i.e. `7fff1fc0033efb0f...` -> `7fff1fc0033ef30f...`. This matches the canonical `ALL_OPERATIONS` / `AVAILABLE_OPERATIONS` lists in the same file, which already exclude 51.

## 2. Unknown/disabled ops were silently hidden and undeletable in the UI

Operations absent from `operationsMap` (such as the disabled ContractType 51) were filtered out of the preview and the edit/delete menus. As a result a user could neither **see** nor **remove** such an op through the interactive UI — making the problem above impossible to self-diagnose or fix manually.

**Fix:** add `opLabel(code)` and `contractTypeName(code)` helpers and render every operation with a visible placeholder (e.g. `Unsupported/Disabled op 51`) instead of dropping it. Four sites updated:
- "Current allowed operations" list in `editActivePermissions` (removed the silent `.filter(...)`)
- the delete list
- `printPermissionSummary` preview
- `printPermissionDetail` final preview

An unknown/disabled op is now visible and deletable.

## Verification

- Decoded both bitmaps: **only** op 51 is removed; no other bits change.
- `./gradlew compileJava` passes.
- No tests pinned the old bitmap string.

## Scope

- Single file changed: `UpdateAccountPermissionInteractive.java`.
- The add-operation menu already sources from `AVAILABLE_OPERATIONS` (which excludes 51), so it was left unchanged.

Fixes #934

### DIFF
--- a/src/main/java/org/tron/core/manager/UpdateAccountPermissionInteractive.java
+++ b/src/main/java/org/tron/core/manager/UpdateAccountPermissionInteractive.java
@@ -86,6 +86,16 @@ public class UpdateAccountPermissionInteractive {
     operationsMap.put("59", "Cancel Unstake");
   }
 
+  private static String opLabel(int code) {
+    String name = operationsMap.get(String.valueOf(code));
+    return name != null ? name : "Unsupported/Disabled op " + code;
+  }
+
+  private static String contractTypeName(int code) {
+    ContractType type = ContractType.forNumber(code);
+    return type != null ? type.name() : "UNKNOWN";
+  }
+
   public String start(String address) {
     System.out.println("\n=== UpdateAccountPermission Interactive Mode ===");
     Response.Account account = WalletApi.queryAccount(WalletApi.decodeFromBase58Check(address));
@@ -105,7 +115,7 @@ public class UpdateAccountPermissionInteractive {
       active.setType(2);
       active.setPermissionName("active");
       active.setThreshold(1L);
-      active.setOperations("7fff1fc0033efb0f000000000000000000000000000000000000000000000000");
+      active.setOperations("7fff1fc0033ef30f000000000000000000000000000000000000000000000000");
       active.setKeys(Lists.newArrayList(new Key(address, 1L)));
       activePermissions = Lists.newArrayList(active);
     }
@@ -443,13 +453,12 @@ public class UpdateAccountPermissionInteractive {
     Collections.sort(currentOps);
 
     while (true) {
-      List<Integer> allowedOps = currentOps.stream()
-          .filter(i -> operationsMap.get(String.valueOf(i)) != null).sorted().collect(Collectors.toList());
+      List<Integer> allowedOps = currentOps.stream().sorted().collect(Collectors.toList());
       System.out.println("\nCurrent allowed operations:");
       for (int i = 0; i < allowedOps.size(); i++) {
         int code = allowedOps.get(i);
-        System.out.println((i + 1) + ". " + operationsMap.get(String.valueOf(code))
-            + " -> " + ContractType.forNumber(code).name() + "(" + code + ")");
+        System.out.println((i + 1) + ". " + opLabel(code)
+            + " -> " + contractTypeName(code) + "(" + code + ")");
       }
 
       System.out.println("\nOperations editing (enter 'q' to finish editing operations):");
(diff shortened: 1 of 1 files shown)


gummy789j:
LGTM. Thanks for the fix.

This addresses the main issue by removing ContractType 51 from the default active-permission bitmap, and it also makes previously hidden/unlisted operations visible and removable in the interactive flow.

I am going to add a follow-up branch/PR to improve the remaining polish: clearer wording for unlisted vs unknown operations, and regression coverage for the default bitmap so ContractType 51 does not get reintroduced.

gummy789j (approved):

PR #950, opened 2026-07-09
feat: ts v0.1.1

gummy789j:
# feat(ts): TypeScript Wallet CLI v0.1.1

Stake/chain queries, governance (vote/reward), a configurable confirmation
timeout, TTY-only secret entry, and a family-keyed command contract — layered
on top of `release_v4.9.8`. User-facing command surface is additive; the one
breaking change is the command-id scheme (see below).

### New commands
- **`vote cast | list | status`** — cast/replace your full SR vote allocation,
  list super representatives & candidates (with brokerage → reward-ratio), and
  show current votes + voting power + reward overview.
- **`reward balance | withdraw`** — claimable voting/block reward and 24h
  withdraw-window status; withdraw accrued rewards.
- **`chain params | prices | node`** — on-chain governance parameters,
  energy/bandwidth unit prices + memo fee, and connected-node sync status.
- **`stake info | delegated`** — staked amounts, voting power, resource usage,
  unfreeze schedule; and V2 delegation records (both directions).
- **`change-password`** — re-encrypt every software keystore under a new master
  password (TTY-only, confirmed).

### Faster, fresher reads (`--wait`)
TRON reads and transaction confirmation now hit the **full node's unconfirmed
view** instead of the solidified node. Confirmation resolves ~one block after
inclusion (~3s) rather than after solidification (~19 blocks / ~60s), and all
reads are fresh and mutually consistent.

### Configurable confirmation cap
New config key **`waitTimeoutMs`** (built-in 60000 ms) supplies the default
`--wait` polling cap; `--wait-timeout` overrides per-invocation.

### Secret policy: wallet secrets are TTY-only
`import mnemonic` / `import private-key` / `change-password` secrets are entered
interactively (hidden prompt) — the `--mnemonic-stdin` / `--private-key-stdin`
channels were removed. Only `password`, `tx`, and `message` remain stdin-backed.
No secret ever sits in argv, env, or the process table.

## Refactors (no behavior change)
- **Family-keyed command contract** — one `ChainCommandDefinition` per logical
  chain command: a service-free `ChainSpec` plus a `family → FamilyBinding`
  table. `registerTronChainCommands` replaces `TronModule`; `commands/tron/*`
  folded up into `commands/*`.
- **`render/index.ts` split** by command domain (wallet / account / tx / vote /
  reward / chain / misc). Golden suite byte-identical.
- **First-class array flags** — repeatable `--for` via yargs `array: true`
  (no preprocess/pipe patch).

## ⚠️ Breaking change
Stable **command id dropped the `tron.` prefix**: chain command ids are now
`path.join(".")` (e.g. `tx.send`, not `tron.tx.send`). The family travels in the
envelope's `chain.family`. The `--json-schema` catalog is otherwise identical
modulo this prefix removal.

## Docs
`ts/docs/typescript-wallet-cli-architecture-source-of-truth.md` updated to match:
command surface, command contract, ports/use-cases, capability list, full-node
read model, secret policy, config keys, and arity.

## Verification
`typecheck` · `depcruise` · `test` · `build` clean; golden suite passing.


fdinesh497-collab (approved):
R

ts/src/adapters/outbound/persistence/fs/index.ts
```diff
@@ -51,6 +51,24 @@ export class AtomicFileStore {
...
+        mkdirSync(dirname(path), { recursive: true });
+        const tmp = `${path}.${process.pid}.${this.#counter++}.tmp`;
+        staged.push({ tmp, path });
+        writeFileSync(tmp, JSON.stringify(value, null, 2) + "\n", { mode: 0o600 });
+      }
+    } catch (e) {
+      for (const { tmp } of staged) { try { unlinkSync(tmp); } catch { /* best-effort */ } }
+      throw e;
+    }
+    for (const { tmp, path } of staged) renameSync(tmp, path); // atomic per file
```
matrix-agent116: **[blocker]** These sequential renames are atomic only per file. If a later `renameSync` fails, earlier keystore blobs have already been replaced while the remaining blobs—and possibly the verifier—still use the old password. `changePassword` then reports a rollback even though the vault may require two passwords or become unusable. Please make the commit recoverable across all files and add a test that injects a failure after one rename succeeds.

ts/src/adapters/outbound/config/index.ts
```diff
@@ -43,6 +44,7 @@ export class ConfigLoader {
       }
       if (raw.defaultOutput === "json" || raw.defaultOutput === "text") defaultOutput = raw.defaultOutput;
       if (typeof raw.timeoutMs === "number") timeoutMs = raw.timeoutMs;
+      if (typeof raw.waitTimeoutMs === "number") waitTimeoutMs = raw.waitTimeoutMs;
```
matrix-agent116: **[suggestion]** Please apply the documented non-negative-integer validation here as well. A manually edited `config.yaml` currently accepts negative or fractional `waitTimeoutMs` values even though `ConfigService` rejects them, allowing an invalid effective configuration.

matrix-agent116 (commented):
## Summary

This PR adds substantial TypeScript CLI functionality and a broad command-contract refactor. However, the new password-rotation path can leave the keystore partially migrated if a rename fails during the multi-file commit, so it is not safe to merge yet.

- **Blocker:** Make password rotation recoverable across the entire set of encrypted blobs and the verifier. The current sequential renames are only atomic per file; add rollback, journaling, or an equivalent recovery mechanism, plus a test that fails after at least one rename has succeeded.
- **Suggestion:** Validate `waitTimeoutMs` when loading `config.yaml`, not only when it is written through `ConfigService`, so manually supplied invalid values cannot enter the effective configuration.

zerodevblock-cyber (approved):

ts/src/adapters/outbound/config/index.ts
gummy789j: Added it.

ts/src/adapters/outbound/persistence/fs/index.ts
gummy789j: Totally agree with this issue. This has been resolved in the latest commit.

ragavan19954 (approved):

PR #955, opened 2026-07-20
security(java): enable fastjson safe mode

gummy789j:
fastjson 1.x parses the "@type" hint even when the target is a JSONObject, and filters it with a deny-list of known gadget classes. wallet-cli feeds untrusted remote payloads into JSON.parseObject (GasFreeApi, MultiSignService, MultiTxWebSocketClient) and never relies on "@type" itself, so force safe mode in Client's static initializer to reject the hint outright. Both the REPL and the standard CLI funnel through Client.main / runMain, so one static block covers both.

Defense in depth rather than a patch for a live vulnerability: no published advisory affects 1.2.83. This matches the default fastjson2 ships with.

Transaction rendering was the compatibility risk, since a Transaction carries its contract in a protobuf Any. TRON's own JsonFormat prints type_url rather than "@type", guarded by TransactionJsonSafeModeTest.

### DIFF
--- a/java/src/main/java/org/tron/common/utils/Utils.java
+++ b/java/src/main/java/org/tron/common/utils/Utils.java
@@ -138,7 +138,7 @@ public class Utils {
 
   public static final int MIN_LENGTH = 2;
   public static final int MAX_LENGTH = 14;
-  public static final String VERSION = " v4.9.6";
+  public static final String VERSION = " v4.9.9";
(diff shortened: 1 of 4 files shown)


zerodevblock-cyber (approved):

albertoblue87-netizen (commented):
Protocolo /955

PR #957, opened 2026-07-23
feat(ts): add tx sign and typed-data sign commands

gummy789j:
Two standalone signing commands so the CLI can sign payloads built elsewhere:

- `tx sign` — sign a transaction constructed outside the CLI, output the signed result for later `tx broadcast`. Pure signer (no owner/contract policy) but enforces payload integrity: txID must be sha256(raw_data_hex), and raw_data must re-encode to those bytes where the contract type is decodable. Verified end-to-end on Nile (confirmed on-chain) and offline.
- `typed-data sign` — EIP-712/TIP-712 in a new top-level `typed-data` group. Byte-for-byte identical output to the agent-wallet SDK for the same key and payload. Verified on a Ledger via Speculos (real TRON app 0.7.6).

Both support software and Ledger accounts. Integrity checking lives in a shared tron/tx-integrity module enforced by both the software strategy and the Ledger adapter, so a hardware account is never the weaker signer. Device signing preliminaries (precheck, awaiting_device, abort-on-timeout) are extracted into obtainSignature and reused by every signing path; the abort signal now actually closes the transport.

Also in this change:
- Map Ledger app-setting APDU errors (0x6a8b/c/d) to actionable messages instead of an opaque UNKNOWN_ERROR.
- Remove the network "required" value: it resolved identically to "optional" (default-network fallback), so it only produced a false help line.
- Fix the Requires help block to state whether a command prompts, and print full signatures in sign-only text receipts instead of a truncated txID.
- Speed up the test suite ~13x by mocking scrypt in keystore-heavy specs (production crypto untouched; crypto.test.ts still uses real scrypt).

### DIFF
--- a/ts/README.md
+++ b/ts/README.md
@@ -7,7 +7,7 @@ The agent-first implementation of wallet-cli, built for automation: every comman
 - **Agent-first** — stable JSON output, deterministic exit codes, and discoverable schemas, built for scripts, CI, and AI agents (details in [The contract, in one paragraph](#the-contract-in-one-paragraph)).
 - **Encrypted local storage** — software keystores are encrypted on disk; secrets are never passed via argv or environment variables.
 - **Software and Ledger signing** — sign in software, or on a Ledger device (the private key never leaves the device).
-- **Covers the main TRON capabilities** — HD wallets, TRX and TRC20/TRC10 transfers, staking / resource delegation, voting / rewards, smart-contract calls and deployment, message signing, and on-chain queries.
+- **Covers the main TRON capabilities** — HD wallets, TRX and TRC20/TRC10 transfers, staking / resource delegation, voting / rewards, smart-contract calls and deployment, message and EIP-712/TIP-712 signing, and on-chain queries.
 
 ## Supported chains
 
@@ -112,6 +112,7 @@ Every command — including every subcommand — has a reference page; run `wall
 | Command | Description |
 |---|---|
 | [`tx send`](docs/commands/tx/send.md) | Send native TRX or TRC20/TRC10 tokens |
+| [`tx sign`](docs/commands/tx/sign.md) | Sign a transaction built elsewhere, without broadcasting |
 | [`tx broadcast`](docs/commands/tx/broadcast.md) | Broadcast a presigned transaction |
 | [`tx status`](docs/commands/tx/status.md) | Show confirmation status (confirmed / failed / pending / not_found) |
 | [`tx info`](docs/commands/tx/info.md) | Show full transaction detail + receipt |
@@ -139,6 +140,7 @@ Every command — including every subcommand — has a reference page; run `wall
 | [`vote`](docs/commands/vote/index.md) | Vote for super representatives ([cast](docs/commands/vote/cast.md) · [list](docs/commands/vote/list.md) · [status](docs/commands/vote/status.md)) |
 | [`reward`](docs/commands/reward/index.md) | Query / withdraw voting rewards ([balance](docs/commands/reward/balance.md) · [withdraw](docs/commands/reward/withdraw.md)) |
 | [`message`](docs/commands/message/index.md) | Sign arbitrary messages ([sign](docs/commands/message/sign.md)) |
+| [`typed-data`](docs/commands/typed-data/index.md) | Sign EIP-712 / TIP-712 structured data ([sign](docs/commands/typed-data/sign.md)) |
 
 ### Local configuration
 
--- a/ts/src/adapters/outbound/chain/tron/signing-strategy.ts
+++ b/ts/src/adapters/outbound/chain/tron/signing-strategy.ts
@@ -5,8 +5,9 @@
  * ADAPTERS table. Adding a chain = one more strategy here.
  */
 import { utils as tronUtils } from "tronweb"
-import type { SignStrategy } from "../../../../domain/types/index.js";
+import type { SignStrategy, TypedDataPayload, TypedDataSignature } from "../../../../domain/types/index.js";
 import { ChainError } from "../../../../domain/errors/index.js"
+import { assertTronTxIntegrity } from "./tx-integrity.js"
 
 // Offline signing via TronWeb's static utils — no TronWeb instance, no fullHost, no network.
 // These are the same primitives the instance methods call internally:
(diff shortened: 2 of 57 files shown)


mrshawon1233-sudo (approved):

Mohammedsapry:
Ok

mrshawon1233-sudo (approved):

mrshawon1233-sudo (approved):

ts/src/adapters/outbound/chain/tron/signing-strategy.ts
```diff
@@ -23,4 +25,17 @@ export const tronSignStrategy: SignStrategy = {
       throw new ChainError("signing_rejected", `TRON message sign failed: ${(e as Error).message}`)
     }
   },
+  async signTypedData(pkHex, payload: TypedDataPayload): Promise<TypedDataSignature> {
+    const { domain, types, message } = payload
+    try {
+      const encoder = tronUtils.typedData.TypedDataEncoder
+      return {
+        signature: tronUtils.typedData.signTypedData(domain as any, types as any, message, pkHex),
```
matrix-agent116, line 33: **[blocker]** The supplied `payload.primaryType` does not affect this signature or the digest: both calls receive only `domain`, the complete `types` map, and `message`; `primaryType` is merely echoed on line 35. With a payload that explicitly selects a declared type other than TronWeb’s inferred root, the command can therefore sign one type while reporting another. Please derive the type set rooted at the selected `primaryType` (or reject an incompatible selection) and use that same selection for signing, digesting, and reporting.

ts/src/adapters/outbound/ledger/index.ts
```diff
@@ -126,25 +163,68 @@ export class Ledger {
...
+    this.assertWired(family);
+    const { domain, types, message } = payload;
+    let digest: string;
+    let primaryType: string;
+    let domainHash: string;
+    let messageHash: string;
+    try {
+      const encoder = tronUtils.typedData.TypedDataEncoder;
+      primaryType = payload.primaryType ?? encoder.from(types as never).primaryType;
+      digest = encoder.hash(domain as never, types as never, message);
```
matrix-agent116, line 223: **[blocker]** This digest is inferred from the complete `types` map, while line 216 computes the device’s struct hash using the explicitly selected `primaryType`. If that selection differs from the inferred root, the returned `digest` does not describe what the Ledger signed. Compute both values from the same rooted type selection and add a test that covers a non-default explicit `primaryType`.

matrix-agent116 (commented):
### Summary

This PR adds valuable standalone transaction and typed-data signing flows with shared device handling and transaction-integrity checks. The overall direction is strong, but I’m requesting changes because an explicitly supplied `primaryType` does not consistently determine the typed data being signed, and the Ledger path can consequently return a digest that differs from the device-signed payload.

### Required improvements

- **Blocker:** Make `primaryType` drive one canonical rooted type set used consistently for the software signature, returned digest, and Ledger struct hash. Please add regression coverage where the supplied `primaryType` differs from the type TronWeb would infer, verifying signature recovery plus the Ledger hashes and returned digest.

zerodevblock-cyber (approved):

ts/src/adapters/outbound/chain/tron/signing-strategy.ts
gummy789j: Fixed in db1013c9. A supplied `primaryType` that is not the root struct (including when referenced through an array field) is now rejected in the domain layer (`domain/typed-data`), so both backends inherit the check and we never sign one type while reporting another. On top of that, `prepareTronTypedData` computes `domainHash`/`messageHash`/`digest` once and asserts `keccak(0x1901 || domainHash || messageHash) === digest`, so any divergence fails closed before signing. Tests: `ts/src/domain/typed-data/index.test.ts` (non-root rejected / root accepted) and `signing-strategy.test.ts` ("honours an explicitly supplied primaryType").

ts/src/adapters/outbound/ledger/index.ts
gummy789j: Fixed in db1013c9. The Ledger path no longer computes hashes on its own — it consumes the shared `prepareTronTypedData` codec, so `domainHash`, `messageHash`, `digest` and `primaryType` all come from one rooted selection and `software.digest === ledger.digest` holds by construction. A non-root `primaryType` is rejected up front in the domain layer, and the codec asserts `keccak(0x1901 || domainHash || messageHash) === digest`, so a mismatch throws before the transport opens (the device is never touched). Tests: `ledger/index.test.ts` ("software and Ledger return the identical digest and primaryType for one payload") and `domain/typed-data/index.test.ts` for the explicit non-default `primaryType` cases.

ts/README.md
```diff
@@ -7,7 +7,7 @@ The agent-first implementation of wallet-cli, built for automation: every comman
 - **Agent-first** — stable JSON output, deterministic exit codes, and discoverable schemas, built for scripts, CI, and AI agents (details in [The contract, in one paragraph](#the-contract-in-one-paragraph)).
 - **Encrypted local storage** — software keystores are encrypted on disk; secrets are never passed via argv or environment variables.
 - **Software and Ledger signing** — sign in software, or on a Ledger device (the private key never leaves the device).
-- **Covers the main TRON capabilities** — HD wallets, TRX and TRC20/TRC10 transfers, staking / resource delegation, voting / rewards, smart-contract calls and deployment, message signing, and on-chain queries.
+- **Covers the main TRON capabilities** — HD wallets, TRX and TRC20/TRC10 transfers, staking / resource delegation, voting / rewards, smart-contract calls and deployment, message and EIP-712/TIP-712 signing, and on-chain queries.
```
kimblekabrain607-wq, line 10: #

kimblekabrain607-wq: Good

ts/README.md
```diff
@@ -112,6 +112,7 @@ Every command — including every subcommand — has a reference page; run `wall
 | Command | Description |
 |---|---|
 | [`tx send`](docs/commands/tx/send.md) | Send native TRX or TRC20/TRC10 tokens |
```
kimblekabrain607-wq, line 114: C l1

ahouramasoud1369-alt (approved):

ts/src/adapters/outbound/chain/tron/signing-strategy.ts
Mohammedsapry: God

ts/src/adapters/outbound/ledger/index.ts
Mohammedsapry: Hi

PR #958, opened 2026-07-23
TS v0.3.0

grayfoxd:


ts/src/application/use-cases/tron/gasfree-service.ts
```diff
@@ -0,0 +1,613 @@
...
+      && totalFee !== undefined
+      && componentTotal !== BigInt(totalFee)
+    ) {
+      throw new ChainError(
+        "gasfree_integrity",
+        "GasFree provider fee components do not match the total fee",
+      );
+    }
+    if (
+      observedTotal !== undefined
```
matrix-agent116, line 587: **[blocker]** Please reject incomplete fee sets before trusting `totalCost`. If a provider supplies `txnTotalCost` (or `estimatedTotalCost`) without both fee components or a corresponding total fee, `observedTotal` is `undefined`, so this condition is skipped and the unverified value is later returned as `totalDeducted`. Require sufficient fields to prove `amount + fees === totalCost` and enforce the authorized `maxFee`; otherwise reject the record as an integrity error.

ts/src/adapters/outbound/persistence/keypair-writer.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (process.platform !== "win32") {
+        const parent = openSync(dirname(target), constants.O_RDONLY);
+        try {
+          fsyncSync(parent);
+        } finally {
+          closeSync(parent);
+        }
+      }
+      return target;
+    } catch (error) {
```
matrix-agent116, line 47: **[blocker]** The destination has already been created before this catch block, but failures from `writeFileSync` or `fsyncSync` only close the descriptor; the partial destination is not removed. That can leave truncated private-key JSON and makes a retry fail with `file_exists`. Please stage and durably write the secret before publishing it with exclusive/no-overwrite semantics, or otherwise guarantee safe cleanup of the file created by this invocation.

ts/src/domain/permission/index.ts
```diff
@@ -0,0 +1,331 @@
...
+  return value;
+}
+
+function permissionName(value: unknown, fallback: string, field: string): string {
+  const name = value === undefined ? fallback : value;
+  if (typeof name !== "string" || name.length === 0) return invalidPermission(`${field} must not be empty`);
+  if (Buffer.byteLength(name, "utf8") > MAX_PERMISSION_NAME_BYTES) {
+    return invalidPermission(`${field} must be at most ${MAX_PERMISSION_NAME_BYTES} UTF-8 bytes`);
+  }
+  if (/\p{Cc}/u.test(name)) return invalidPermission(`${field} must not contain control characters`);
```
matrix-agent116, line 114: **[suggestion]** Please reject Unicode format controls (`\p{Cf}`) here as well as control characters. Permission names are rendered directly in terminal views, and the central sanitizer only strips C0/C1 bytes, so bidi controls such as U+202E can visually reorder or disguise permission information. The account/contact validators already treat format controls as unsafe.

matrix-agent116 (commented):
### Summary

This is a substantial release adding GasFree transfers, local and TronLink multisig workflows, permission management, contacts, encoding utilities, and account lifecycle commands. The overall design includes strong validation, but two failure paths should be fixed before merging because they can accept inconsistent provider data or leave an unusable private-key artifact.

### Required improvements

- **Blocker:** Strengthen GasFree fee validation so `estimatedTotalCost`/`txnTotalCost` cannot appear without enough fee data to verify both `amount + fees == totalCost` and the signed `maxFee` bound. Reject incomplete or inconsistent fee sets instead of skipping validation.
- **Blocker:** Make keypair publication failure-atomic. A write or `fsync` failure currently leaves the exclusively created destination in place, potentially containing a partial JSON/private key and preventing a safe retry.
- **Suggestion:** Reject Unicode format controls in permission names. These names are rendered in terminal permission views, while the output sanitizer only removes C0/C1 control bytes, so bidi format characters can produce misleading text.

PR #963, opened 2026-07-28
Feat/ts v4.11.0

gummy789j:
## Summary

Brings the TypeScript CLI to the 4.11.0 release line: **TRON multi-signature support end to end**, **gas-free token transfers**, and local utilities (contacts, keypair generation, encoding conversion, account lifecycle, receive QR).

Command surface grows **52 → 66**, and all 66 now have a reference page.

Scope is `ts/` only, apart from a one-line version constant in Java.

**Base:** `release_v4.11.0` · 15 commits · 159 files, +12,991 / −216

## Multi-signature

The CLI can now participate in a multi-sig account for the whole lifecycle: configure permissions, collect signatures from several keys, broadcast once the threshold is met.

| Command | Purpose |
|---|---|
| `permission show` / `update` | Inspect groups with decoded operation bitmaps; replace the full structure |
| `tx sign --hex/--file` | **Append** a signature, preserving prior ones (`--check` verifies online) |
| `tx approvals` | Who signed, accumulated weight, missing weight, expiry |
| `tx multisig` | Collect signatures via the TronLink service (`--create` / `--sign` / `--watch`) |
| `tx broadcast --hex/--file` | Validates threshold and expiry before submitting |

Signatures travel either **offline** as a hex artifact passed between signers, or through the **TronLink service** as a shared queue.

Safety properties worth review:

- `tx sign` on an artifact is offline by default; permission membership is only claimed with `--check`.
- `tx broadcast` refuses a transaction below threshold (`not_authorized`) or expired (`tx_expired`) — it
- `permission update` rejects rather than normalizes (`invalid_permission`), emits lockout warnings, and re-reads after confirmation. A bitmap cannot silently widen a permission: unnamed set bits must be declared in                      `unknownOperationIds`.
- TronLink responses are untrusted — the tx hash is re-derived and contract type, owner, weights, and signature progress cross-checked (`provider_error`). `--watch` receives a count only, never content.                                   
## GasFree                                                                                                                                                                                                                                   
`gasfree info` / `transfer` / `trace` — move USDT and other supported tokens with **no TRX**, by signing a TIP-712 `PermitTransfer` and letting the provider broadcast for a fee in the transferred token.                                   
Balance and fees are checked up front; the signed digest is recomputed and the recovered signer verified against the selected account (`signing_rejected`); provider metadata is validated (`gasfree_integrity`). `--dry-run` prices without unlocking, `--wait` reports settled rather than estimated amounts. Mainnet and Nile only; Shasta returns

## Local utilities

| Command | Purpose |
|---|---|
| `contact add` / `list` / `remove` | Recipient names usable as `tx send --to` and `gasfree transfer --to
| `address generate` | Keypair generated locally, **not** added to the wallet |
| `encoding convert` | base58 ⇄ `41…` hex ⇄ `0x` EVM, and hex / Base64 / Base58Check |
| `account activate` / `set` | Activate an account; set the one-time on-chain name or ID |
| `current --qr` | Terminal receive QR for the selected address |

- The address book is re-validated on every read (regular file, owner, mode `0600`); names resembling an to` can't be reinterpreted.
- `address generate` writes the key to an exclusively created `0600` file and never overwrites; stdout requires explicit `--print-secret`.
- `encoding convert` refuses any 32-byte input — indistinguishable from a private key, and argv is visibl
- `current --qr` encodes exactly the address and never draws a partial QR (a truncated code could scan as a different address).

## Architecture & config

New ports keep hexagonal boundaries intact: `contact-repository`, `gasfree-provider`, `keypair-writer`, `qr-encoder`, `tronlink-collaboration`. New pure-domain modules: `contact`, `encoding`, `gasfree`, `permission`.

Five new optional config keys — `tronlinkSecretId` / `SecretKey` / `Channel` and `gasfreeApiKey` / `ApiSecret`. Both secrets are **masked on read and write** (`********`, including the echoed `input`), so a config write is safe to log;
unset keys are omitted entirely.

## Documentation

19 new pages, 9 updated, following the existing layout. New groups `permission/`, `gasfree/`, `contact/`,new pages `tx/approvals`, `tx/multisig`, `account/activate`, `account/set`; updated `tx/index` (multi-siglifecycle), `tx/broadcast`, `tx/send`, `current`, `config`, `README`, and the agent `SKILL.md`.

Written against the command registry and domain rules with examples taken from real output. Verified mechanically: all 66 commands have a page, and every relative link and anchor resolves.

## Version

`0.2.0` → `4.11.0` for the `ts` package, aligning it with the Java release line. Java `Utils.VERSION` moves `v4.10.0` → `v4.11.0`.

## Verification

```
tsc --noEmit     clean
vitest run       98 files, 675 passed, 2 skipped
depcruise src    no violations (303 modules, 817 dependencies)
```

41 test files added or updated.

## Behaviour changes

- `tx send --to` accepts a contact name; resolution is surfaced in the receipt and `data.toContact`.
- `tx send` auth is now `conditional` — `--build-only` builds unsigned without unlocking.
- `tx broadcast` validates before submitting, so some transactions now fail locally instead of on-chain.
- `current` honours `--account`, so `data.active` is no longer always `true`.
- Signing commands gain `--build-only`, `--permission-id`, `--expiration`.


ts/src/application/use-cases/tron/contract-service.ts
```diff
@@ -89,6 +98,8 @@ export class TronContractService {
       account: scope.activeAccount,
       broadcaster: gateway,
       ...transactionMode(input),
+      ...tronTransactionHooks(gateway),
```
zerodevblock-cyber: **[P1][CR-020] Recompute the deployment address after transaction preparation**

`tronTransactionHooks()` runs after `build()`. A non-zero permission ID or a custom expiration changes `raw_data` and recomputes the txID, but `contractAddress` is still the value captured from the builder transaction. TronWeb derives the contract address from the owner and txID, so the CLI can report a different address from the contract that is actually deployed.

Please derive the address from the final prepared transaction, or have the pipeline return the final derived address. A regression test should cover non-zero permission IDs and custom expirations.

ts/docs/commands/config.md
```diff
@@ -77,13 +99,39 @@ wallet-cli config timeoutMs 120000 -o json
 {"schema":"wallet-cli.result.v1","success":true,"command":"config","data":{"key":"timeoutMs","value":120000,"input":"120000"},"meta":{"durationMs":3,"warnings":[]}}
 ```
 
+Configure a service credential — note that the value never appears in the output:
+
+```bash
+wallet-cli config gasfreeApiSecret "$SECRET" -o json
```
zerodevblock-cyber: **[P1][CR-016] Do not pass service credentials through process arguments**

This example expands `$SECRET` before the process starts, so the credential is present in OS argv even though the command masks its output. Process inspection, endpoint telemetry, CI tracing, or command logging can capture it; this also contradicts the machine-interface promise that secrets never travel via argv.

Please provide a dedicated stdin, hidden TTY, or file-descriptor secret input and update this example to use that channel.

ts/src/application/use-cases/tron/account-service.ts
```diff
@@ -58,8 +71,154 @@ export class TronAccountService {
...
+      account: scope.activeAccount,
+      broadcaster: gateway,
+      ...mode,
+      ...tronTransactionHooks(gateway),
+      confirm: tronConfirmation(gateway, scope),
+      build: (owner) => gateway.buildAccountCreate(owner, input.address),
+      estimate: async () => ({ ...fee, balanceSun }),
+    });
+    if (outcome.stage === "confirmed" && !outcome.failed) {
+      const activated = await gateway.getAccount(input.address);
```
zerodevblock-cyber: **[P2][CR-002] Preserve the confirmed receipt when this post-check cannot be read**

At this point the transaction has already returned `confirmed`. If this follow-up RPC times out, is rate-limited, or observes a lagging node, the exception escapes the use case and replaces the confirmed txid with a command failure. Callers can then retry an operation that already succeeded.

Please retain the confirmed/submitted outcome and report an observation warning for read failures. The account-set and permission-update post-checks have the same partial-success boundary and should follow the same rule.

ts/src/application/use-cases/tron/gasfree-service.ts
```diff
@@ -0,0 +1,613 @@
...
+    timeoutMs: number,
+  ): Promise<GasFreeTransferRecord | undefined> {
+    const deadline = this.clock() + Math.max(0, timeoutMs);
+    let previous = initial.state;
+    if (previous === "SUCCEED" || previous === "FAILED") return initial;
+    for (;;) {
+      const remaining = deadline - this.clock();
+      if (remaining <= 0) return undefined;
+      await this.sleep(Math.min(POLL_MS, remaining));
+      const current = await this.provider.trace(network, initial.id);
```
zerodevblock-cyber: **[P2][CR-003] Preserve the accepted GasFree receipt when polling fails**

`submitTransfer()` has already succeeded and the trace ID has been captured before this request runs. A transient timeout, 429, or provider 5xx escapes from here, causing the entire command to fail without returning that trace ID. Only normal deadline exhaustion currently preserves the submitted receipt.

Please retry retryable polling failures within the wait deadline, or return the existing `accepted` result with a warning so callers can reconcile the original submission instead of resubmitting it.

ts/src/adapters/inbound/cli/contracts/envelope.ts
```diff
@@ -12,7 +15,7 @@ export interface ChainView {
 }
 export interface Meta {
   durationMs: number;
-  warnings: string[];
+  warnings: WarningItem[];
```
zerodevblock-cyber, line 18: **[P2][CR-007] Keep the v1 warning element type stable**

The envelope still declares `wallet-cli.result.v1`, while the published machine contract defines `meta.warnings` as `string[]`. Allowing `WarningView` here serializes `{code, message}` objects for permission warnings and can break existing v1 consumers that parse, join, or display every element as a string.

Please keep `warnings` as strings in v1, or introduce the structured form through a versioned field/schema with an explicit migration path.

ts/src/adapters/outbound/chain/tron/transaction-codec.ts
```diff
@@ -0,0 +1,224 @@
...
+  33: "UpdateSettingContract",
+  41: "ExchangeCreateContract",
+  42: "ExchangeInjectContract",
+  43: "ExchangeWithdrawContract",
+  44: "ExchangeTransactionContract",
+  45: "UpdateEnergyLimitContract",
+  46: "AccountPermissionUpdateContract",
+  48: "ClearABIContract",
+  49: "UpdateBrokerageContract",
+  52: "MarketSellAssetContract",
```
zerodevblock-cyber, line 56: **[P2][CR-008] Add the missing ShieldedTransferContract codec mapping**

This table jumps from type 49 to 52, while the permission catalog and public signing documentation identify protocol type 51 as `ShieldedTransferContract`. Because every hex/file sign, approval, broadcast, and collaboration path passes through this codec, a valid type-51 artifact is rejected as unsupported.

Please add a lossless type-51 codec mapping and round-trip fixture, or narrow the published support contract if this transaction type is intentionally unsupported.

ts/src/adapters/outbound/config/index.ts
```diff
@@ -35,10 +35,21 @@ export class ConfigLoader {
...
+    let gasfreeApiSecret: string | undefined;
 
     const path = ConfigLoader.configPath(env);
     if (existsSync(path)) {
       const raw = parseYaml(readFileSync(path, "utf8")) ?? {};
+      if (
+        (typeof raw.tronlinkSecretKey === "string" && raw.tronlinkSecretKey !== "")
+        || (typeof raw.gasfreeApiSecret === "string" && raw.gasfreeApiSecret !== "")
+      ) {
+        assertSecretConfigPermissions(path);
```
zerodevblock-cyber, line 51: **[P2][CR-021] Route this bootstrap error through the standard formatter**

This guard is correct to fail closed, but `ConfigLoader.load()` runs inside `composeCliRuntime()` before the runner's `try` block and before the formatter exists. As a result, even `-o json --help` emits a raw `fatal:` line and exits 1 instead of producing the v1 error envelope and the `UsageError` exit code 2.

Please bring configuration/bootstrap failures under a minimal normalized error boundary while preserving this permission check and redacting sensitive parser context.

ts/src/adapters/outbound/persistence/keypair-writer.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (process.platform !== "win32") {
+        const parent = openSync(dirname(target), constants.O_RDONLY);
+        try {
+          fsyncSync(parent);
+        } finally {
+          closeSync(parent);
+        }
+      }
+      return target;
+    } catch (error) {
```
zerodevblock-cyber, line 62: **[P3][CR-005] Do not leave a partially written final keypair file on failure**

The final path has already been created before this catch. If `fchmod`, write, file fsync, close, or directory fsync fails, the handler closes the descriptor but leaves an empty, truncated, or durability-unknown file at the final path. `O_EXCL` then blocks a retry using the same path.

Please write to an exclusive same-directory temporary file and atomically publish it after fsync, or reliably remove an unpublished final file on every failure path. Fault-injection tests should cover each stage.

ts/src/adapters/inbound/cli/commands/tx.ts
```diff
@@ -2,18 +2,24 @@ import { z } from "zod";
 import type { ChainSpec, FamilyBinding } from "../contracts/index.js";
 import { UsageError } from "../../../../domain/errors/index.js";
 import type { TronTransactionService } from "../../../../application/use-cases/tron/transaction-service.js";
+import type { TronSigService } from "../../../../application/use-cases/tron/sig-service.js";
+import type { TronMultisigService } from "../../../../application/use-cases/tron/multisig-service.js";
+import type { TronMultisigCollaborationService } from "../../../../application/use-cases/tron/multisig-collaboration-service.js";
+import type { TransactionArtifactWriter } from "../../../outbound/persistence/transaction-artifact-writer.js";
```
zerodevblock-cyber: **[P3][CR-009] Keep the inbound adapter independent of the persistence implementation**

This creates an inbound-CLI → concrete-outbound dependency, and the command later orchestrates `writer.write()` itself. That makes the signing workflow harder to reuse or test with another storage implementation; the current dependency rule also misses this type-only reverse edge.

Please define an application-layer artifact-writer port or save use case and inject its implementation from the composition root. It would also be useful to add an architecture check that includes type-only imports.

zerodevblock-cyber (commented):
## Review summary

I validated the 15 selected findings against PR #963 at commit `1e068529da4bf2535618acb228fa5597da5c49de`. Each finding was rechecked against the aggregate diff, its source-to-sink call path, existing guards, tests, and a safe local or in-memory reproduction where feasible. The detailed findings are attached as inline comments.

### Findings

- **P1 (2)**
  - `CR-020`: transaction preparation can change the deployment txID while the CLI keeps reporting the address derived from the pre-prepare txID.
  - `CR-016`: documented service-secret configuration passes the expanded credential through OS argv before output masking applies.
- **P2 (5)**
  - `CR-002`: post-confirmation read failures overwrite an irreversible success result.
  - `CR-003`: a polling error after GasFree acceptance discards the trace ID and submitted receipt.
  - `CR-007`: `wallet-cli.result.v1` changes warning elements from strings to objects without versioning.
  - `CR-008`: the transaction codec rejects protocol type 51 despite the published support surface.
  - `CR-021`: insecure-config bootstrap failures bypass the JSON envelope and UsageError exit code.
- **P3 (8)**
  - `CR-005`, `CR-009`, `CR-010`, `CR-012`, `CR-013`, `CR-014`, `CR-015`, and `CR-022` cover file-write atomicity, dependency direction, error classification, documentation drift, machine-error consistency, GasFree failure semantics, terminal bidi safety, and TronLink signer metadata validation.

The highest-priority fixes are to bind the reported deployment address to the final prepared transaction (`CR-020`) and move service secrets off argv (`CR-016`). The P2 items should also be resolved before release because they affect irreversible-operation recovery, machine compatibility, supported transaction types, and deterministic CLI error handling.

Validation checks completed:

- the PR head and reviewed HEAD are identical;
- all 15 inline anchors are added lines in the current aggregate PR diff;
- the relevant targeted suite passed (6 files, 50 tests);
- `npm run typecheck` passed;
- no external service was called, no real credential was used, and no transaction was signed or broadcast.

ts/src/adapters/outbound/chain/tron/tron.ts
```diff
@@ -161,19 +436,20 @@ export class TronRpcClient implements TronGateway, Broadcaster {
 
   async getTokenInfo(contract: string): Promise<TronTokenInfo> {
     return this.#wrap("trc20 tokenInfo", async () => {
-      const c = await this.#tw.contract().at(contract);
+      // Read the view methods by selector rather than via contract().at(): tokens deployed without a
+      // published ABI (e.g. USDD on Nile) resolve to a contract object with no methods at all.
+      const read = async (fn: string): Promise<string | undefined> =>
+        this.#constant(contract, fn, []).then(([hex]) => hex).catch(() => undefined);
```
zerodevblock-cyber: **[P3][CR-010] Preserve transport errors for strict metadata validation**

This blanket catch converts timeout, 429, connection, and node failures into missing metadata. The GasFree path then interprets those missing fields as `gasfree_integrity`, so callers cannot distinguish a retryable node outage from an actual provider/token mismatch.

Please provide a strict metadata-read path that preserves typed transport failures, and only downgrade failures that are explicitly known to mean an optional method is unavailable.

ts/src/adapters/inbound/cli/commands/shared.ts
```diff
@@ -10,10 +10,13 @@ import { TextFormatters } from "../render/index.js";
...
 
 // ── execution-mode flags shared by every signing command ─────────────────────────
-/** dry-run / sign-only fields; default (no flag) = sign AND broadcast on-chain. */
+/** Transaction execution fields; default (no mode flag) = sign and broadcast on-chain. */
 export const txModeFields = {
-  dryRun: z.boolean().default(false).describe("build and estimate only, with no signature and no broadcast; mutually exclusive with --sign-only"),
-  signOnly: z.boolean().default(false).describe("sign and output the transaction without broadcasting; mutually exclusive with --dry-run; broadcast later with tx broadcast"),
+  dryRun: z.boolean().default(false).describe("build and estimate only, with no signature and no broadcast"),
+  signOnly: z.boolean().default(false).describe("sign and output complete transaction hex without broadcasting"),
+  buildOnly: z.boolean().default(false).describe("build and output unsigned complete transaction hex without unlocking"),
```
zerodevblock-cyber: **[P3][CR-012] Update every leaf command page for the new transaction modes**

`--build-only`, `--permission-id`, and `--expiration` are now shared by 11 transaction commands, but their individual reference pages still document only the previous dry-run/sign-only surface; some also still imply that every non-dry-run mode unlocks the wallet.

Please update the affected leaf pages and consider validating each page's option set against the command registry in CI so shared schema changes cannot leave command documentation behind.

ts/src/adapters/outbound/persistence/keypair-writer.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (descriptor !== undefined) closeSync(descriptor);
+      const code = (error as NodeJS.ErrnoException).code;
+      if (code === "EEXIST" || code === "ELOOP") {
+        throw new ExecutionError(
+          "file_exists",
```
zerodevblock-cyber: **[P3][CR-013] Reuse the existing output-conflict error contract**

An existing target is reported here as `file_exists` via `ExecutionError` (exit 1), while the existing backup writer reports the equivalent `O_EXCL` conflict as `output_exists` via `UsageError` (exit 2). Automation can therefore treat the same deterministic conflict as retryable in one command and malformed input in another.

Please use one stable code, error class, and exit-code policy for exclusive-output conflicts, with a compatibility plan if the older code must remain supported.

ts/src/application/use-cases/tron/gasfree-service.ts
```diff
@@ -0,0 +1,613 @@
...
+      return accepted;
+    }
+    const settledAmount = terminal.txnAmount ?? accepted.amount;
+    const settledServiceFee =
+      terminal.txnTransferFee ?? accepted.serviceFee;
+    const settledActivateFee =
+      terminal.txnActivateFee ?? accepted.activateFee;
+    return {
+      ...accepted,
+      stage: terminal.state === "SUCCEED" ? "confirmed" : "failed",
```
zerodevblock-cyber, line 285: **[P3][CR-014] Do not expose a FAILED transfer as top-level success**

This branch resolves normally with `stage: "failed"`, so the shared shell formats it as `success: true` and exits 0. The published machine interface tells automation to branch on exit code/top-level success first and does not define a GasFree-specific secondary check, so a failed transfer can be recorded as successful.

Please either return a failure envelope/non-zero exit for the FAILED terminal state, or explicitly version and document a stable two-level status contract. Add an end-to-end assertion for exit code, top-level success, stage, state, and failure reason.

ts/src/adapters/inbound/cli/render/permission.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+  if (current) lines.push(current);
+  return lines;
+}
+
+function permissionCard(permission: PermissionGroupView, active = false): string {
+  const activePermission = active ? permission as AccountPermissionsView["actives"][number] : undefined;
+  const lines = [
+    `Permission Name   ${permission.name}  (id ${permission.id}${active ? ", active" : ""})`,
```
zerodevblock-cyber, line 25: **[P3][CR-015] Escape Unicode bidi controls in untrusted display fields**

Permission names are chain-controlled and reach this security-sensitive terminal summary. The shared text sanitizer removes ANSI/C0/C1 bytes but leaves Unicode bidi controls such as U+202E and U+2066–U+2069, allowing the visible order of names, IDs, or adjacent labels to differ from the underlying string. GasFree and multisig renderers expose the same class of input.

Please escape or visibly mark bidi/default-ignorable controls in text mode while preserving the original value in JSON, and add terminal golden tests for the affected renderers.

ts/src/application/use-cases/tron/multisig-collaboration-service.ts
```diff
@@ -0,0 +1,418 @@
...
+    if (
+      approval.txId !== remote.view.txId
+      || approval.contractType !== remote.view.contractType
+      || approval.currentWeight !== remote.view.currentWeight
+      || approval.permission.id !== remote.view.permission.id
+      || approval.permission.threshold !== remote.view.permission.threshold
+      || approval.signatures !== remote.view.signatures
+      || approval.expiration !== remote.view.expiration
+      || approved.size !== signedProgress.length
+      || signedProgress.some((address) => !approved.has(address))
```
zerodevblock-cyber, line 327: **[P3][CR-022] Validate every progress entry against the on-chain permission**

This comparison only binds the signed-address set to the chain-approved set. Unsigned progress entries and all per-signer weights remain provider-controlled, so TronLink can add the current account as an unsigned signer or redistribute individual weights while preserving the signed set and aggregate `currentWeight`. The list can then report a false `awaitingMySignature` state or incorrect weights.

Please compare every progress address and weight with the selected permission's on-chain keys and render chain-derived weights. `signChecked()` still prevents unauthorized signing, but the collaboration list should not present unverified signer metadata as chain-validated.

ts/src/application/use-cases/tron/contract-service.ts
gummy789j: Fixed. `refreshTransactionIdentity` re-derives `contract_address` (it is part of the transaction identity, not an independent field), and `deploy` captures it from the prepared transaction instead of the builder output.

One scope correction: the default path was never wrong — TronWeb omits `Permission_id` unless it is truthy, so `delete` is a no-op and the txID is unchanged. Only `--permission-id N≠0` and `--expiration` were affected, and `--permission-id` is legal in broadcast mode, so this did reach chain. Regression tests added; the old fake pipeline never invoked `prepare`, which is why the suite was blind to it.

ts/docs/commands/config.md
gummy789j: Not changing this, deliberately.

`SecretKind` is a closed set — `password | privateKey | mnemonic | tx | message` — pinned by architecture doc §11 and by the channel list in the machine interface's "Secret handling" section. Service credentials are config values, not `SecretKind` secrets, so "Secrets never travel via argv" is not contradicted; those docs are simply silent about them.

What remains true is that argv persists into shell history and CI logs. That is an accepted trade-off for a local single-user CLI.

ts/src/application/use-cases/tron/account-service.ts
gummy789j: Fixed for all three sites, via a shared `warnOnPostCheck`: both the unreadable-read and the mismatch paths degrade to a warning (`<code>_unavailable` / `<code>_mismatch`) so the confirmed txid always survives.

One correction — `permission update` was already half-right: it degraded a mismatch to `permission_postcheck_mismatch` rather than throwing. That code and message are preserved verbatim; what it lacked was protection for a `show()` transport failure.

ts/src/application/use-cases/tron/gasfree-service.ts
gummy789j: Fixed. Transport flakiness (`TransportError`, or code `timeout`) retries inside the existing deadline; any other non-integrity error warns and returns `submitted` with the traceId intact. The classification already existed in the GasFree client, so no new mechanism was needed.

Went one step further: the integrity errors were also discarding the trace id — exactly when you most need it — so all three now carry it in `error.details`.

ts/src/adapters/inbound/cli/contracts/envelope.ts
gummy789j: Keeping the union and correcting the contract instead of flattening.

`owner_lockout` is the only stable machine-readable signal that a permission update may lock the wallet out; flattening would leave only unstable English prose to match on. `machine-interface.md` now declares `(string | {code, message})[]` with a section on reading it, and states that `code` values are stable within v1 while warning `message` text is not.

Also fixed three related mismatches found while checking this — `vote/status.md` documented a `zero_reward_ratio` code in three places that does not exist anywhere in `src/`.

ts/src/adapters/outbound/chain/tron/transaction-codec.ts
gummy789j: Confirmed, but the scope is off in both directions. Asking TronWeb's deserializer by type name (its "not supported" error is payload-independent), **four** types cannot be decoded: `UnfreezeAssetContract` (14), `ShieldedTransferContract` (51), `MarketSellAssetContract` (52), `MarketCancelOrderContract` (53). The doc named three and missed 14, and said they were "still bound by check 1" — they are refused outright, not weakly checked.

A lossless type-51 codec is not available to build (TronWeb has no shielded support), so 51 is now in the table purely so the refusal names the type instead of an opaque id, and the doc lists all four as refused.

ts/src/adapters/outbound/config/index.ts
gummy789j: Fixed. `composeCliRuntime()` runs in its own boundary now, reporting through a formatter that depends on argv alone — the config that would supply the default output mode is exactly what failed.

Worth noting the 0600 check is the narrowest trigger: it early-returns on Windows. The platform-independent ones matter more — a malformed `config.yaml` printed the offending line verbatim, i.e. potentially the credential beside it. `ConfigLoader` now classifies read/parse failure as `invalid_config` (UsageError) with a fixed message, so a config typo is no longer a generic `internal_error`.

ts/src/adapters/outbound/persistence/keypair-writer.ts
gummy789j: Fixed. The catch now unlinks the file, but only when `O_EXCL` is what created it — unlinking on the `EEXIST` path would destroy the user's pre-existing file, so there is a test pinning that a pre-existing target survives a failed write byte-for-byte.

Chose unlink-on-failure over temp-file-plus-rename: there is no prior version to keep intact here (overwriting is forbidden outright), so the only property needed is "do not leave a roadblock".

ts/src/adapters/inbound/cli/commands/tx.ts
gummy789j: Fixed with a real port: `application/ports/transaction-artifact-writer.ts`, implementation renamed to `SecureTransactionArtifactWriter` (matching the existing `SecureKeypairWriter` / `KeypairWriter` pair), composition root wires it.

On the architecture check — `tsPreCompilationDeps: true` does make depcruise see type-only edges, but it also closes four pre-existing type-only cycles through the domain barrels and `price/`, failing `no-circular` for reasons unrelated to any boundary. `dependencyTypesNot` filters one edge, not a whole cycle. The invariant is pinned by `adapters/inbound/boundary.test.ts` instead; depcruise config untouched.

ts/src/adapters/outbound/chain/tron/tron.ts
gummy789j: Fixed by deleting the catch rather than classifying around it.

Measured against a live Nile node: a view method the contract does not implement returns `result: true` with `constant_result: [""]`, so `#constant` never throws for it and it decodes to undefined unaided. The catch only ever swallowed a transport fault and "no contract at this address". The existing test's stub returned `{result: {result: false}}` for a missing method — that inaccurate model is what made the catch look load-bearing; it now models the four measured shapes.

Blast radius was wider than reported: `tx send --contract` was reporting a retryable node timeout as a `UsageError` at exit 2.

ts/src/adapters/inbound/cli/commands/shared.ts
gummy789j: Confirmed and larger than reported. Taking the CLI's own `--json-schema` catalog as the source of truth: **14** commands carry the three flags, and **11** of their pages documented none of them — all six `stake` pages among them.

The password point understates it too: those pages claim the master password unconditionally, which is wrong for `--build-only` **and** `--dry-run` — neither unlocks. Fixed across the 11 pages, and added the registry-vs-docs check you suggested (a page whose options table has a `--sign-only` row must document all three).

ts/src/adapters/outbound/persistence/keypair-writer.ts
gummy789j: Aligned to the existing contract: `output_exists` / `UsageError` / exit 2, identical to `SecureBackupWriter`.

No compatibility plan needed — `address generate` is new in this PR, so `file_exists` was never released; it now appears nowhere in `src/` or `docs/`. Worth noting the two findings compounded: a leftover file (CR-005) made the retry hit `file_exists`, which was classified exit 1 "safe to retry" — a script would have retried forever.

ts/src/application/use-cases/tron/gasfree-service.ts
gummy789j: Documented; behaviour unchanged for now.

One correction: this is not GasFree-specific. `pipeline.ts` returns `stage: "failed"` the same way, so `tx send --wait` on a reverted transaction is also `success: true` at exit 0. Making only GasFree exit non-zero would leave the two inconsistent, which is worse than the current state.

The real gap was that the contract never wrote the rule down — "Script safety" spelled out the four-state model for `tx status` but said nothing about what a `--wait` receipt means. That is now stated in `machine-interface.md`, `tx/send.md` and `gasfree/transfer.md`. Changing the exit code is still on the table as one change covering both commands, and it needs the txId to survive into the error envelope first.

ts/src/adapters/inbound/cli/render/permission.ts
gummy789j: Fixed. `sanitizeText` is applied once to the whole text frame in `HumanOutputFormatter`, so the single seam covers the permission, GasFree and multisig renderers plus error lines and progress events, without touching any renderer.

Bidi and default-ignorable characters are **marked, not stripped** (`Treasury<U+202E> ecnanetniam`): dropping them would silently mangle legitimate right-to-left text and leave no trace anything was removed. Ordinary Arabic/Hebrew text is unaffected (asserted), and JSON stays byte-exact.

ts/src/application/use-cases/tron/multisig-collaboration-service.ts
gummy789j: Fixed. `#verifyOnChain` now reads `getSignWeight` — the same call `create()` already uses for this check — requires the selected account and every reported signer to be a key in the permission, and renders weights from the chain. Signed weights are re-summed from chain values, closing redistribution among the signed keys.

On the threat model: injection is not easy (endpoint pinned in the builtin networks, HTTPS enforced, `networks` read-only), so this needs the service itself to be compromised. The reason to fix is the non-adversarial case — `signature_progress` is a snapshot, so a `permission update` while a transaction sits pending leaves it stale, and stale data was rendering as chain-validated.

One check deliberately not added: requiring every on-chain key to appear. I cannot verify the live service always returns the full roster, and every figure the user acts on is chain-derived anyway.

ts/docs/commands/tx/sign.md
```diff
@@ -98,6 +118,21 @@ echo "$PW" | wallet-cli tx sign --transaction "$TX" --password-stdin -o json \
 wallet-cli tx broadcast --network tron:nile --tx-stdin < signed.json
 ```
 
+Append a signature without an approval RPC dependency, then check it from an online machine:
+
+```bash
+echo "$PW" | wallet-cli tx sign --file partially-signed.hex \
+  --out signed.hex --password-stdin
```
zerodevblock-cyber: P2: This example is described as running without an approval RPC dependency, but it does not pass `--offline`. For `--file`/`--hex` input, `offline` defaults to false and the binding invokes `signChecked`, which queries signer authorization and approval state from the node. The documented offline workflow will therefore fail in the environment it claims to support. Please add `--offline` here (or relabel this as an online checked-signing flow) and align the broader `tx sign` documentation with the input-specific behavior.

ts/src/adapters/inbound/cli/commands/tx.ts
```diff
@@ -82,38 +121,163 @@ export const txBroadcastTronBinding = (svc: TronTransactionService): FamilyBindi
...
+  description:
+    "With no mode flag, list service-managed transactions for the selected account. --create signs\n" +
+    "an UNSIGNED transaction locally and submits it, which opens the collection at the first\n" +
+    "signature; --sign fetches the accumulated transaction, signs locally, and submits it;\n" +
+    "--watch opens the official WebSocket count-only notification channel.",
+  requires: [
+    "TronLink service credentials — config tronlinkSecretId / tronlinkSecretKey / tronlinkChannel",
+  ],
+  baseFields: tronLinkMultisigFields,
+  baseRefine: tronLinkMultisigRefine,
```
zerodevblock-cyber, line 250: P2: `tronLinkMultisigRefine` rejects combinations of `--create`, `--sign`, and `--watch`, but this spec does not declare the corresponding `exclusive` metadata. The new help renderer and JSON catalog only project constraints from `spec.exclusive`, so users and agents discover these modes as independently optional until execution fails. Please add an `at-most-one` exclusive group for the three modes and cover the real command in help/catalog regression tests.

zerodevblock-cyber (commented):
Review summary: 2 P2 findings. Both concern CLI contract discoverability and documentation consistency.

ts/src/adapters/inbound/cli/commands/tx.ts
gummy789j: Verified fixed in 37496ace. `txTronLinkMultisigSpec` now declares an `at-most-one` exclusive group for `create`, `sign`, and `watch`. Regression coverage uses the real command spec and verifies both human help rendering and the JSON catalog projection. Targeted help tests pass.

ts/docs/commands/tx/sign.md
gummy789j: Verified fixed in 37496ace. The default `--file`/`--hex` example is now explicitly described as the online checked-signing path, and the no-node-access workflow includes `--offline` before checking approvals later from an online machine. The documentation now matches the input-specific runtime behavior. Targeted CLI tests pass.

zerodevblock-cyber (approved):

octavio12345300 (approved):

Aranmobinarann:
diff --git a/ts/src/adapters/inbound/cli/output/redact.ts b/ts/src/adapters/inbound/cli/output/redact.ts
new file mode 100644
index 0000000..1111111
--- /dev/null
+++ b/ts/src/adapters/inbound/cli/output/redact.ts
@@ -0,0 +1,175 @@
+/**
+ * Redaction helpers for CLI output.
+ *
+ * Rules (per team choices):
+ * - txId/txid are whitelisted (kept).
+ * - Long hex-like strings (0x?[0-9a-fA-F]{40,} ) -> "[redacted]".
+ * - Keys that look sensitive (privateKey, secret, mnemonic, password, signature(s), hex, raw_data_hex, signed, signedTx)
+ *   are replaced with "[redacted]" in JSON outputs.
+ * - Uint8Array values are replaced with "[redacted]".
+ * - Opt-in to include sensitive fields via env WALLET_CLI_INCLUDE_SENSITIVE=1.
+ *
+ * Provide:
+ * - redactSensitive(obj) -> deep-copy with redactions (for JSON envelopes)
+ * - redactText(s) -> sanitize user-facing text output
+ */
+const SENSITIVE_KEY_PATTERNS = [
+  /^privateKey$/i,
+  /^private_key$/i,
+  /^secret$/i,
+  /^mnemonic$/i,
+  /^password$/i,
+  /^signature(s)?$/i,
+  /^signed$/i,
+  /^signedTx$/i,
+  /^signed_tx$/i,
+  /^hex$/i,
+  /^raw_data_hex$/i,
+  /^rawDataHex$/i,
+  /^tx$/i,
+];
+
+const LONG_HEX_RE = /\b0x?[0-9a-fA-F]{40,}\b/g; // threshold 40
+const TXID_RE = /\b[0-9a-fA-F]{64}\b/; // txid pattern (whitelisted)
+const INCLUDE_SENSITIVE = process.env.WALLET_CLI_INCLUDE_SENSITIVE === "1";
+
+/** Whether a property name is considered sensitive. */
+export function isSensitiveKey(key: string | number | symbol | null): boolean {
+  if (key === null) return false;
+  const name = String(key);
+  return SENSITIVE_KEY_PATTERNS.some((re) => re.test(name));
+}
+
+/** Deep-redact an arbitrary value for safe JSON output. Returns a brand-new value. */
+export function redactSensitive<T>(input: T): T {
+  if (INCLUDE_SENSITIVE) return input;
+
+  const seen = new WeakSet<object>();
+
+  function walk(value: unknown, keyName: string | null = null): unknown {
+    if (value === null || value === undefined) return value;
+
+    // Primitives
+    if (typeof value === "string") {
+      // txId (64 hex) is permitted; keep as-is if it looks like txid and not long hex blob
+      if (TXID_RE.test(value) && !LONG_HEX_RE.test(value)) return value;
+      // redact long hex-like runs
+      if (LONG_HEX_RE.test(value)) return "[redacted]";
+      return value;
+    }
+    if (typeof value === "number" || typeof value === "boolean" || typeof value === "bigint") return value;
+
+    // Binary data
+    if (value instanceof Uint8Array) return "[redacted]";
+
+    // Arrays
+    if (Array.isArray(value)) {
+      return value.map((v) => walk(v, keyName));
+    }
+
+    // Objects
+    if (typeof value === "object") {
+      if (seen.has(value as object)) return "[redacted-cycle]";
+      seen.add(value as object);
+
+      const out: Record<string, unknown> = {};
+      for (const [k, v] of Object.entries(value as Record<string, unknown>)) {
+        if (isSensitiveKey(k)) {
+          out[k] = "[redacted]";
+          continue;
+        }
+        // If the property value is a long-hex string, redact it
+        if (typeof v === "string" && LONG_HEX_RE.test(v)) {
+          out[k] = "[redacted]";
+          continue;
+        }
+        // txId/txid whitelist: keep
+        if ((k.toLowerCase() === "txid" || k.toLowerCase() === "tx_id") && typeof v === "string") {
+          out[k] = v;
+          continue;
+        }
+        out[k] = walk(v, k);
+      }
+      return out;
+    }
+
+    return "[redacted]";
+  }
+
+  return walk(input) as T;
+}
+
+/** Sanitize free-form text output (human-facing) by redact long hex runs and sensitive tokens.
+ *  Use when rendering text to terminal when WALLET_CLI_INCLUDE_SENSITIVE is not set.
+ */
+export function redactText(s: string): string {
+  if (INCLUDE_SENSITIVE) return s;
+  if (!s) return s;
+  // Replace long hex runs with [redacted], but keep txId-like 64-hex when it sits alone (common case).
+  return s.replace(LONG_HEX_RE, (match) => {
+    // Allow txid per decision: if exactly 64 hex (no 0x) and not prefixed by 0x, keep it.
+    const cleaned = match.replace(/^0x/, "");
+    if (/^[0-9a-fA-F]{64}$/.test(cleaned)) {
+      return match;
+    }
+    return "[redacted]";
+  });
+}
diff --git a/ts/src/adapters/inbound/cli/output/envelope.ts b/ts/src/adapters/inbound/cli/output/envelope.ts
index 3a8a6d7..2222222 100644
--- a/ts/src/adapters/inbound/cli/output/envelope.ts
+++ b/ts/src/adapters/inbound/cli/output/envelope.ts
@@ -1,68 +1,83 @@
 /**
  * OutputEnvelope — the result/error envelope builder for the OutputFormatter. Shapes
  * the user-facing `wallet-cli.result.v1` contract: schema version, chain view, and meta.
  * Pure (no I/O); the formatter turns the envelope into strings.
  */
-import type { NetworkDescriptor } from "../../../../domain/types/index.js";
-import type { ChainView, ErrorEnvelope, Meta, ResultEnvelope, WarningItem } from "../contracts/index.js";
-
-type CliErrorEnvelopeShape = { code: string; message: string; details?: object };
-
-const SCHEMA_VERSION = "wallet-cli.result.v1" as const;
-
-/** JSON serialization that keeps big numbers as strings. */
-export function toJson(value: unknown): string {
-  return JSON.stringify(value, (_k, v) => {
-    if (typeof v === "bigint") return v.toString();
-    if (v instanceof Uint8Array) return Buffer.from(v).toString("hex");
-    return v;
-  });
-}
-
-function chainView(net: NetworkDescriptor): ChainView {
-  return {
-    family: net.family,
-    network: net.id,
-    chainId: net.chainId,
-  };
-}
-
-function meta(durationMs: number, warnings: WarningItem[]): Meta {
-  return { durationMs, warnings };
-}
-
-export const OutputEnvelope = {
-  success(
-    command: string,
-    net: NetworkDescriptor | undefined,
-    data: unknown,
-    m: { durationMs: number; warnings: WarningItem[] },
-  ): ResultEnvelope {
-    const env: ResultEnvelope = {
-      schema: SCHEMA_VERSION,
-      success: true,
-      command,
-      data: data ?? {},
-      meta: meta(m.durationMs, m.warnings),
-    };
-    if (net) env.chain = chainView(net); // neutral commands omit chain
-    return env;
-  },
-
-  error(
-    command: string,
-    net: NetworkDescriptor | undefined,
-    err: CliErrorEnvelopeShape,
-    m: { durationMs: number; warnings: WarningItem[] },
-  ): ErrorEnvelope {
-    const env: ErrorEnvelope = {
-      schema: SCHEMA_VERSION,
-      success: false,
-      command,
-      error: err,
-      meta: meta(m.durationMs, m.warnings),
-    };
-    if (net) env.chain = chainView(net);
-    return env;
-  },
-};
+import type { NetworkDescriptor } from "../../../../domain/types/index.js";
+import type { ChainView, ErrorEnvelope, Meta, ResultEnvelope, WarningItem } from "../contracts/index.js";
+import { redactSensitive, redactText } from "./redact.js";
+
+type CliErrorEnvelopeShape = { code: string; message: string; details?: object };
+
+const SCHEMA_VERSION = "wallet-cli.result.v1" as const;
+
+/** JSON serialization that keeps big numbers as strings. */
+export function toJson(value: unknown): string {
+  // redactSensitive is a no-op if WALLET_CLI_INCLUDE_SENSITIVE=1
+  const safe = redactSensitive(value);
+  return JSON.stringify(safe, (_k, v) => {
+    if (typeof v === "bigint") return v.toString();
+    if (v instanceof Uint8Array) return Buffer.from(v).toString("hex"); // after redactSensitive, Uint8Array values are redacted
+    return v;
+  });
+}
+
+function chainView(net: NetworkDescriptor): ChainView {
+  return {
+    family: net.family,
+    network: net.id,
+    chainId: net.chainId,
+  };
+}
+
+function meta(durationMs: number, warnings: WarningItem[]): Meta {
+  return { durationMs, warnings };
+}
+
+export const OutputEnvelope = {
+  success(
+    command: string,
+    net: NetworkDescriptor | undefined,
+    data: unknown,
+    m: { durationMs: number; warnings: WarningItem[] },
+  ): ResultEnvelope {
+    const env: ResultEnvelope = {
+      schema: SCHEMA_VERSION,
+      success: true,
+      command,
+      data: data ?? {},
+      meta: meta(m.durationMs, m.warnings),
+    };
+    if (net) env.chain = chainView(net); // neutral commands omit chain
+    return env;
+  },
+
+  error(
+    command: string,
+    net: NetworkDescriptor | undefined,
+    err: CliErrorEnvelopeShape,
+    m: { durationMs: number; warnings: WarningItem[] },
+  ): ErrorEnvelope {
+    const env: ErrorEnvelope = {
+      schema: SCHEMA_VERSION,
+      success: false,
+      command,
+      error: err,
+      meta: meta(m.durationMs, m.warnings),
+    };
+    if (net) env.chain = chainView(net);
+    return env;
+  },
+};
+
+// Re-export redactText so renderers can call it before emitting human-facing text.
+export { redactText };
diff --git a/ts/src/adapters/inbound/cli/output/stream-wrapper.ts b/ts/src/adapters/inbound/cli/output/stream-wrapper.ts
new file mode 100644
index 0000000..2222222
--- /dev/null
+++ b/ts/src/adapters/inbound/cli/output/stream-wrapper.ts
@@ -0,0 +1,55 @@
+// ts/src/adapters/inbound/cli/output/stream-wrapper.ts
+import type { StreamManager } from "../contracts/runtime.js";
+import { redactText } from "./redact.js";
+
+/**
+ * wrapStreamManager — decorate a StreamManager so that all human-facing text
+ * is passed through redactText by default (unless WALLET_CLI_INCLUDE_SENSITIVE=1).
+ *
+ * Use: const wrapped = wrapStreamManager(originalStream); pass `wrapped` to the CLI runner.
+ */
+export function wrapStreamManager(stream: StreamManager): StreamManager {
+  return {
+    result(text: string): void {
+      try {
+        stream.result(redactText(String(text)));
+      } catch (e) {
+        // fall back to original to avoid hiding errors in the output path
+        stream.result(text);
+      }
+    },
+    diagnostic(level, msg) {
+      // diagnostics are structured; ensure any human strings are redacted
+      const safe = { ...msg } as any;
+      if (typeof safe.message === "string") safe.message = redactText(safe.message);
+      stream.diagnostic(level, safe);
+    },
+    errorLine(msg: string) {
+      try {
+        stream.errorLine(redactText(String(msg)));
+      } catch (e) {
+        stream.errorLine(msg);
+      }
+    },
+    event(frame: string | null) {
+      if (frame === null) return stream.event(null);
+      try {
+        stream.event(redactText(String(frame)));
+      } catch (e) {
+        stream.event(frame);
+      }
+    },
+    readStdinOnce: stream.readStdinOnce.bind(stream),
+    warnings: stream.warnings.bind(stream),
+  };
+}
diff --git a/ts/src/adapters/inbound/cli/output/__tests__/stream-wrapper.test.ts b/ts/src/adapters/inbound/cli/output/__tests__/stream-wrapper.test.ts
new file mode 100644
index 0000000..3333333
--- /dev/null
+++ b/ts/src/adapters/inbound/cli/output/__tests__/stream-wrapper.test.ts
@@ -0,0 +1,38 @@
+import { wrapStreamManager } from "../stream-wrapper.js";
+
+test("wrapStreamManager redacts long hex in result", () => {
+  let last = "";
+  const fakeStream = {
+    result: (t: string) => { last = t; },
+    diagnostic: (_l: any, _m: any) => {},
+    errorLine: (_m: string) => {},
+    event: (_f: string | null) => {},
+    readStdinOnce: () => "",
+    warnings: () => [],
+  };
+  const wrapped = wrapStreamManager(fakeStream as any);
+  const hex = "0x" + "ab".repeat(40); // long hex
+  wrapped.result(`here is secret ${hex} end`);
+  expect(last).toContain("[redacted]");
+  expect(last).not.toContain("ab".repeat(40));
+});
 
###

ghost (commented):
C

PR #971, opened 2026-08-05
Release v4.11.0

gummy789j:
### Change

1. **Multi-sig is now a first-class workflow in the TypeScript CLI.** `permission show` and `permission update` read and replace an account's permission structure on chain (the latter burns 100 TRX and warns before an owner lockout); every broadcast command gained `--permission-id <n>`, `--expiration <ms>`, and a new `--build-only` early-exit mode alongside `--dry-run` / `--sign-only`. `tx sign` now takes a transaction hex via `--hex` / `--file`, appends exactly one signature while preserving prior ones, and reports how far the accumulated weight is from the threshold — refusing before a key is ever decrypted if the account is not in the permission group (`not_authorized`) or has already approved (`already_signed`); `--offline` keeps the air-gapped path. `tx approvals` is the read-only companion that shows the permission group, threshold, accumulated weight, approved signers, and expiration without signing. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

2. **`tx multisig` adds optional collaboration through the TronLink multi-sig service.** Where the on-chain path passes a hex from person to person, the service holds the transaction, accumulates signatures, and pushes notifications: the default mode lists transactions awaiting this account, `--create` signs an unsigned hex and opens a collection at `1 / N`, `--sign <txId>` co-signs a pending one, and `--watch` keeps a WebSocket open and reports only the count awaiting your signature (never transaction content). Once the threshold is reached the service broadcasts. Credentials (`tronlinkSecretId` / `tronlinkSecretKey` / `tronlinkChannel`) are set with `config` and are per-environment; without them the command fails with `tronlink_credentials_missing`. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

3. **`gasfree` brings gas-less token transfers to the TypeScript CLI.** `gasfree info` reports your GasFree address, activation status, nonce, and fee schedule; `gasfree transfer` signs an EIP-712 structured-data transfer and submits it to the GasFree provider, which puts it on chain and charges the fee in the transferred token itself, so no TRX is required; `gasfree trace <traceId>` follows the provider's `WAITING → INPROGRESS → CONFIRMING → SUCCEED / FAILED` states. Because submission goes to a provider rather than a node, the receipt carries a `traceId` instead of a `txId` — follow it with `--wait` or `gasfree trace`, not `tx status`. Credentials are `gasfreeApiKey` / `gasfreeApiSecret` via `config`. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

4. **New account-lifecycle and local-utility commands.** `account activate` creates a not-yet-existing address on chain without transferring any asset, with the payer covering the creation fee; `account set` writes the on-chain name or account id (permanent on mainnet, so it is deliberately one-at-a-time). A local contact book — `contact add` / `list` / `remove`, stored `0600` in the config directory — lets a saved name be used wherever a recipient is expected (`tx send --to`, `gasfree transfer --to`). `encoding convert <input>` auto-detects an input and prints every equivalent representation (TRON base58 / TRON hex / EVM / public key, or hex / Base64 / Base58Check), rejecting secrets outright. `address generate` produces a throwaway keypair offline, writing the private key to a `0600` file rather than the terminal unless `--print-secret` is passed. `current --qr` renders a scannable receive-address QR in text mode without unlocking or touching the network. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

5. **The Java documentation was split out of `java/README.md` into a browsable `java/docs/` tree.** The README shrank from a single ~2,700-line page to an overview with quick links; commands now live under `docs/commands/` by domain (wallet, account, transfers, staking, multisig, GasFree, DEX, proposals, contracts, …), with `docs/concepts/`, `docs/guide/`, and a field-by-field `docs/reference/config.md`. Stale planning and QA scratch documents were removed. No Java behavior changed in this release apart from the version constant. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

### Change

1. **A transaction the node rejected could be reported as submitted.** TronWeb does not throw on rejection — it hands back the node's response verbatim, and a rejected `/wallet/broadcasttransaction` carries no `result` field at all, so the previous `result === false` check never fired and the CLI returned a success envelope with a txId for a transaction that was never accepted. Acceptance is now white-listed (`result !== true` is a rejection) on both the object and `--hex` broadcast paths, surfacing the node's reason as `transaction_rejected`. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

2. **`block` lost precision on large int64 fields.** Block responses were parsed through JavaScript numbers, so protobuf int64/uint64 values beyond `Number.MAX_SAFE_INTEGER` came back rounded. Blocks are now fetched and parsed losslessly, with out-of-range integers preserved as exact decimal strings. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

3. **`--dry-run` reported doomed staking requests as fine.** `stake freeze` and `stake unfreeze` are rejected by the node at broadcast time when the amount exceeds the available balance or the amount actually staked for that resource, but `--dry-run` never reaches the node — so a request that could not possibly succeed previewed cleanly. Both now pre-flight against the chain and fail with `insufficient_balance` / `insufficient_stake`. Relatedly, `contract send` now warns when the supplied `--fee-limit` is below the energy estimate at the current energy price, and the staked-amount computation was corrected for the node's habit of omitting the default `BANDWIDTH` enum in `frozenV2`. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

4. **Chain-controlled text could repaint the terminal.** Permission names, token names and symbols, and co-signer labels are written by whoever put them on chain. Terminal control bytes were already stripped, but bidirectional and zero-width formatting characters passed straight through — `U+202E` reverses the display order of everything after it, letting a crafted name change how the address or weight beside it appears, and zero-width characters can make two different names render identically. Text mode now escapes them visibly (`<U+202E>`) rather than dropping them, so legitimate right-to-left text still displays normally and tampering is obvious. JSON output is never rewritten. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

5. **Filesystem failures surfaced as raw OS errors.** `backup` and the atomic config/keystore writers let Node `ErrnoException`s escape, producing `internal_error` and, on a failed write, leaving a truncated or empty file at the destination. Failures are now typed as `io_error` with the partial artifact removed, and the multi-file atomic write reports a rolled-back failure distinctly from one that committed but could not confirm durability. Two new config-file conditions are also reported explicitly: `invalid_config` when `config.yaml` cannot be parsed (the parser detail is withheld, since it quotes the offending line and may carry a credential) and `insecure_config` when the file holds service credentials but is a symlink or group/world-readable. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))

6. **Positional arguments rejected number-shaped values.** Yargs inferred a type for the anonymous positional tail, so a value such as `0xdeadbeef` or `12345` arrived at its string-typed field as a number and was rejected. The tail is now typed as strings so the raw token survives parsing. As a consequence, a field exposed as a positional no longer also accepts its `--<field>` spelling — this affects the undocumented `config --key/--value` and `block --number` forms; the documented positional form is unchanged, and the global `--account` flag still works on `use` / `rename` / `delete` / `backup`. ([#963](https://github.com/tronprotocol/wallet-cli/pull/963))


matrix-agent116 (approved):

zerodevblock-cyber (approved):

jj5419952-stack:
Ok

PR #984, opened 2026-08-24
Code Smells for evm-extension

grayfoxd:


ts/src/adapters/inbound/cli/commands/contract.ts
```diff
@@ -637,11 +626,6 @@ export const contractDeploySpec: ChainSpec = {
   description:
     "Deploy contract creation bytecode and report the new contract's address.\n" +
     "Flags marked (tron) or (evm) apply only on networks of that family; using one on the other family is rejected.",
-  // The Ledger TRON app firmware rejects CreateSmartContract (APDU 0x6a80), even with
-  // blind-signing enabled; software accounts sign and deploy it fine.
-  requires: [
-    "a software (non-Ledger) account (tron) — the Ledger TRON app cannot sign a contract deployment; the Ledger Ethereum app can",
-  ],
   baseFields: deployFields,
```
gummy789j: Please restore the `requires` entry that was removed just above this line.

`help-review-v4.13.0.md` **C7** (lines 33 and 268) asks for the opposite of removal: keep the line and tag it `(tron)`, because the limitation is Ledger *TRON* app firmware — the Ledger Ethereum app signs a deploy fine. The text already carried that tag.

The "neutralize shared metadata" rationale is fair — `FamilyBinding` has no `requires` field, so there is no per-family home for it today. But deleting it trades a documented constraint for silence. If we want it family-scoped, adding `requires?: string[]` to `FamilyBinding` is worth doing as separate work.

ts/src/adapters/inbound/cli/commands/typed-data.ts
```diff
@@ -21,9 +21,9 @@ export const typedDataSignSpec: ChainSpec = {
   capability: "typedData.sign",
   summary: "Sign EIP-712 / TIP-712 structured data",
   description:
-    "Sign an EIP-712 / TIP-712 typed-data payload with the selected account.\n" +
-    "`EIP712Domain` in `types` is ignored, `value` is accepted for `message`, and TRON base58\n" +
-    "addresses work in address fields.",
+    "Prints the signature, the digest that was signed, and the primary type.\n" +
```
gummy789j: This restores a string that `help-review-v4.13.0.md:256` flags as defective — non-imperative and describing output rather than the action, violating §10.1 rules 1 and 2. The current `Sign an EIP-712 / TIP-712 typed-data payload with the selected account.` is the fix for it.

Also, line 26 replaces the concrete "TRON base58 addresses work in address fields" with "interpreted by the selected chain family's signing strategy". That is true but vaguer, and "signing strategy" is an internal term leaking into user-facing help. The base58 support is real (`tron/signing-strategy.ts` goes through `tronUtils.typedData`), so the concrete sentence is worth keeping.

ts/src/adapters/inbound/cli/commands/tx.ts
```diff
@@ -41,7 +41,7 @@ export const txSendSpec: ChainSpec = {
   auth: "conditional",
   broadcasts: true,
   capability: "tx.send",
-  summary: "Send the native coin or a token",
+  summary: "Send native coins or tokens with human --amount",
```
gummy789j: `help-review-v4.13.0.md:233` prescribes exactly the current text, `Send the native coin or a token`; line 253 flags the `…with human --amount` tail as the defect being fixed. It also reads oddly now that `--raw-amount` exists alongside `--amount`. Could we keep the current summary?

ts/src/adapters/outbound/chain/evm/evm.ts
```diff
@@ -186,7 +187,13 @@ export class EvmRpcClient implements EvmGateway {
       ...(price === undefined ? {} : { effectiveGasPriceWei: price.toString(10) }),
       ...(r.blockNumber === undefined
         ? {}
-        : { blockNumber: Number(BigInt(String(r.blockNumber))) }),
+        : {
+            blockNumber: quantityToSafeNumber(
```
gummy789j: This is now stricter than its two neighbours: `gasUsed` and `effectiveGasPrice` a few lines up still go through the tolerant `BigInt(String(...))`, and the old code here was `Number(BigInt(String(r.blockNumber)))` — the explicit `String()` was load-bearing.

A provider that returns `blockNumber: 123` as a number now fails the **entire** receipt with `rpc_error`, breaking `tx status` / `tx info`. Same for `blockNumber: null`, which `contractAddress` right below explicitly guards against but this does not.

Suggest `quantityToSafeNumber(String(r.blockNumber), …)` plus a `!== null` check in the guard.

ts/src/adapters/outbound/chain/evm/evm.ts
```diff
@@ -196,6 +203,10 @@ export class EvmRpcClient implements EvmGateway {
 
   /** the JSON-RPC envelope, unthrown — callers that classify errors themselves need to see it. */
   async #send(method: string, params: unknown[]): Promise<JsonRpcResponse> {
```
gummy789j: `#send` is now a one-line pass-through to `#request` with a single caller (line 145) — a redundant indirection in a code-smell PR. Simplest fix: rename `#request` to `#send` and drop the wrapper.

ts/src/domain/types/tx.ts
```diff
@@ -283,7 +283,7 @@ export interface TxInfoView extends TxParties {
   /** head height minus this transaction's block; best-effort, see TxStatusView.confirmations. */
   confirmations?: number;
   energyUsed?: number; // tron execution resource
-  gasUsed?: number; // evm execution resource
+  gasUsed?: number | string; // evm execution resource
```
gummy789j: Open question rather than a change request. The direction is right — it aligns `tx info` with `evm-confirmation.ts`, which already emitted a decimal string. But a union type makes agents write `typeof` branches, and this is a JSON contract change (`21000` → `"21000"`). Since EVM is not public yet, this is the moment to commit to `string` outright rather than leave the union. Worth noting TRON's sibling `energyUsed` stays `number`, so the two families would disagree on shape.

gummy789j (commented):
The extractions (`tx-build.ts`), the safe-integer helpers, the malformed-JSON guard, the required-dependency cleanup and the family-scoped `--json-schema` fix are all good and I'd like to take them.

PR #990, opened 2026-08-27
feat: v4.13.0 (evm support & java standardCLI deprecate)

gummy789j:


ts/src/adapters/outbound/persistence/migration.ts
```diff
@@ -0,0 +1,96 @@
...
+   * orphaning its encrypted key blob.
+   */
+  private write(stale: StaleFile[], password: string | undefined, held = 0): void {
+    if (held === stale.length) return this.commit(stale, password);
+    this.store.withLock(stale[held]!.step.path, () => this.write(stale, password, held + 1));
+  }
+
+  private commit(stale: StaleFile[], password: string | undefined): void {
+    this.store.writeJsonAll(
+      stale.flatMap(({ step, doc, storedVersion }) => [
```
zerodevblock-cyber: **CR-001 (P1)**

The migration commits `doc`, which was read during planning before these locks were acquired. If another process migrates the same files and then adds or changes an account or token while this process is waiting for consent or a password, this write replaces the newer document with the stale snapshot and silently loses that state. Please re-read and re-plan under the complete lock set, or reject the commit when a revision/hash check shows that an input changed.

ts/src/application/use-cases/evm/transaction-service.ts
```diff
@@ -0,0 +1,610 @@
...
+   * a bare "not found" invites the reader to conclude it never did.
+   */
+  async status(
+    scope: TransactionScope,
+    network: NetworkDescriptor,
+    hash: string,
+  ): Promise<TxStatusView> {
+    const gateway = this.gateways.get(network, "evm");
+    const [transaction, receipt, head] = await Promise.all([
+      gateway.getTransactionByHash(hash).catch(() => null),
```
zerodevblock-cyber: **CR-002 (P2)**

This converts every transaction lookup failure—including timeouts, rate limits, server errors, and JSON-RPC errors—into the same `null` value used for a successful “not found” response. As a result, `tx status` can return `not_found` or `pending` with exit code 0 when the node never answered. Only an explicit RPC `result: null` should represent absence; transport and RPC failures need to remain errors.

ts/src/adapters/outbound/chain/evm/signing-strategy.ts
```diff
@@ -0,0 +1,100 @@
...
+  async signTypedData(pkHex, payload: TypedDataPayload): Promise<TypedDataSignature> {
+    const { domain, types, message } = payload;
+    // A JSON-RPC eth_signTypedData payload carries EIP712Domain in `types`, but ethers derives
+    // the domain separator from `domain` itself and rejects the redundant entry. Dropping it is
+    // what lets the wallet accept the payload shape dApps actually send.
+    const structTypes = Object.fromEntries(
+      Object.entries(types as Record<string, unknown>).filter(([name]) => name !== "EIP712Domain"),
+    ) as Record<string, Array<{ name: string; type: string }>>;
+    try {
+      const digest = TypedDataEncoder.hash(domain as never, structTypes, message);
```
zerodevblock-cyber, line 87: **CR-004 (P2)**

The declared `primaryType` does not select the type dependency closure used for this digest. A valid payload with `primaryType: "Mail"` and an unrelated extra type passes normalization, but all types are passed to ethers here, which rejects the payload as having ambiguous roots. Build the recursive type closure from `primaryType` and use the same normalized set for both software and Ledger signing.

ts/src/adapters/inbound/cli/commands/exchange.ts
```diff
@@ -172,6 +173,7 @@ export const exchangeTradeSpec: ChainSpec = {
       .optional()
       .describe("derive the floor from current reserves, less this percentage"),
```
zerodevblock-cyber, line 174: **CR-005 (P2)**

Rounding the percentage to basis points can loosen the user's minimum-receive protection. For example, `1.006%` slippage with a predicted return of `1,000,000` produces `989,900`, while the exact floor is `989,940`, allowing execution below the requested threshold. Preserve the decimal input for exact integer arithmetic, or reject precision finer than the supported two decimal places.

ts/.gitignore
```diff
@@ -7,4 +7,5 @@ dist/
 docs/superpowers
 docs/qa
 docs/adr
```
zerodevblock-cyber, line 9: **CR-012 (P3)**

Production code added references to ADRs for migration ordering, serialization, pricing, Ledger constraints, and alias behavior, but `docs/adr` is ignored and no referenced ADR is present in the repository. Maintainers therefore cannot retrieve the rationale that the comments rely on to distinguish protocol invariants from workarounds or product decisions. The referenced ADRs need a versioned, reachable location.

ts/src/application/use-cases/wallet-service.ts
```diff
@@ -184,22 +205,29 @@ export class WalletService {
...
+    const privateKey = this.#exportablePrivateKey(account, family);
     const file = this.backups.write(
       descriptor.accountId,
       requestedPath,
-      KeystoreV3.encrypt(
-        privateKey,
-        masterPassword,
-        tronHexAddress(descriptor.addresses[KEYSTORE_FAMILY]!),
-      ),
+      KeystoreV3.encrypt(privateKey, masterPassword, keystoreAddress(family, privateKey)),
```
zerodevblock-cyber, line 228: **CR-013 (P3)**

Failures from `randomBytes` or `randomUUID` escape this domain operation without translation, so `backup --keystore` reports `internal_error` even though the public error catalog defines `entropy_failure`. This leaves the stable entropy error unreachable and prevents callers from classifying the failure correctly. Route entropy/ID generation through an injectable boundary and map provider failures to `entropy_failure`.

ts/src/domain/migration/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  needsPassword: boolean;
+}
+
+export interface MigrationPlan {
+  stale: MigrationCandidate[];
+  needsPassword: boolean;
+}
+
+export function planMigrations(candidates: MigrationCandidate[]): MigrationPlan {
+  const stale = candidates.filter((c) => c.storedVersion < c.currentVersion);
```
zerodevblock-cyber, line 22: **CR-015 (P2)**

A stored version greater than `currentVersion` is treated exactly like the current schema: it produces no migration plan and startup continues. The current CLI can then mutate and rewrite a future wallet document using older semantics, potentially violating fields or invariants it does not understand. Startup and repository writes should fail closed when `storedVersion > currentVersion`.

ts/src/adapters/inbound/cli/commands/exchange.ts
```diff
@@ -172,6 +173,7 @@ export const exchangeTradeSpec: ChainSpec = {
       .optional()
```
zerodevblock-cyber, line 173: **CR-019 (P1)**

Returning `UNPROTECTED_FLOOR` here lets the default broadcast path sign a trade with `expected = 1`, so any non-zero output is accepted regardless of price deterioration. A warning does not prevent an irreversible asset operation, especially for automated callers. Broadcasting should require an economically meaningful floor or an explicit unsafe opt-out with confirmation before signing.

ts/src/adapters/inbound/cli/commands/wallet.ts
```diff
@@ -259,14 +277,12 @@ export function registerWalletCommands(
...
-    summary: "Import an account from a standard Web3 keystore JSON",
+    summary: "Import a Web3 keystore file",
     description:
-      "Import a single account from a standard Web3 keystore JSON (as exported by TronLink or\n" +
-      "'backup --keystore'), stored encrypted under your master password and made active. It carries\n" +
-      "one private key, so nothing can be derived from it; a same-address account is refused with\n" +
-      "account_exists (delete it first).\n\n" +
-      "Interactive-only: the master password and the keystore's own password are entered only via\n" +
-      "hidden TTY prompts, never stdin/argv — without a TTY it fails with tty_required.",
+      "Import a single account from a standard Web3 keystore JSON (TronLink, `backup --keystore`).\n" +
```
zerodevblock-cyber, line 282: **CR-020 (P2)**

The imported file controls `dklen` and the PBKDF2/scrypt cost parameters, but validation only checks integer shape and `dklen >= 32`. A user-selected malicious or corrupted keystore can therefore force excessive synchronous CPU or memory work before MAC verification, causing long hangs, `RangeError`, or OOM. Enforce compatible parameter bounds and an aggregate resource budget before deriving the key.

ts/src/adapters/outbound/config/index.ts
```diff
@@ -70,11 +76,39 @@ export class ConfigLoader {
...
+      if (raw.aliases && typeof raw.aliases === "object" && !Array.isArray(raw.aliases)) {
+        for (const [alias, target] of Object.entries(raw.aliases as Record<string, unknown>)) {
+          if (typeof target === "string") aliases[alias.toLowerCase()] = target;
+        }
+        // A target may itself be an alias — most often a legacy id the book still forwards.
+        // Resolution is single-hop, so fold the target here rather than leave an entry resolve()
+        // cannot follow. One hop only: the book is hand-written, and a chain of two is a mistake
+        // to surface, not a feature to support.
+        for (const [alias, target] of Object.entries(aliases)) {
+          const forwarded = aliases[target.toLowerCase()];
```
zerodevblock-cyber: **CR-023 (P1)**

This forwarding pass treats canonical network IDs as ordinary alias keys. If a user defines an alias named `tron:3448148188` that points to mainnet, the built-in `nile -> tron:3448148188` entry is rewritten to mainnet as well, so selecting `nile` can build, sign, and broadcast on the wrong chain. Canonical IDs must be reserved terminal targets and must not participate as user-defined forwarding aliases.

java/docs/reference/config.md
```diff
@@ -111,7 +111,9 @@ tronlink = {
 
 ## Connecting to Java-tron
 
-wallet-cli connects to Java-tron via the gRPC protocol, which can be deployed locally or remotely. Configure the Java-tron node IP and port in `src/main/resources/config.conf` so wallet-cli can talk to the node. You can also use `SwitchNetwork` to switch among mainnet, testnets (Nile and Shasta), and custom networks — see [commands/network](../commands/network.md).
+wallet-cli connects to Java-tron via gRPC. The startup network is inferred by comparing `fullnode.ip.list` and `soliditynode.ip.list` with the built-in Mainnet, Nile, and Shasta endpoints; any other pair is `CUSTOM`. Consequently, `net.type = mainnet` with Nile endpoints still starts on Nile. Check the endpoints themselves before sending funds.
```
zerodevblock-cyber: **CR-024 (P2)**

Filling a missing Solidity endpoint with the fullnode endpoint changes the values later used for network identification. For a documented fullnode-only Nile configuration, the two values become identical and cannot match Nile's distinct fullnode/Solidity endpoint pair, so the known network is classified as `CUSTOM`. Preserve which endpoint was originally absent and identify the network from the endpoints actually supplied, rather than from fallback-mutated values.

ts/src/adapters/outbound/tokenbook/builtins.test.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, it, expect } from "vitest";
+import { OFFICIAL_TOKENS } from "./builtins.js";
+import { isEvmAddress } from "../../../domain/address/index.js";
+
+// These are constants nobody re-derives at runtime: `tx send --token USDT` trusts the book's
+// contract AND decimals without asking the chain. A mistyped address is therefore a fund-loss
+// shape — and EIP-55 is what catches it, since altering one character breaks the checksum.
+describe("official EVM token entries", () => {
+  const evmNetworks = Object.entries(OFFICIAL_TOKENS).filter(([id]) => id.startsWith("evm:"));
```
zerodevblock-cyber: **CR-026 (P2)**

Production EVM token keys now use the `eip155:` namespace, but this filter still searches for `evm:`. `evmNetworks` is therefore empty, causing both parameterized validations and the duplicate check to pass without examining any official EVM token entry. Enumerate the current EVM family/namespace and assert that the selected set is non-empty and covers every production EVM token key.

zerodevblock-cyber (changes requested):
This review confirmed 12 issues: 3 P1, 7 P2, and 2 P3.

P1:

- CR-001: Migration commits an outdated snapshot read before acquiring the lock, potentially overwriting concurrent wallet or token state updates.
- CR-019: When no execution protection is specified, exchange defaults `expected` to `1`, allowing execution for any non-zero return.
- CR-023: Canonical IDs can reverse-pollute built-in aliases, causing a testnet name to resolve to mainnet.

P2:

- CR-002: EVM status masks RPC failures as successful `not_found` or `pending` results.
- CR-004: EIP-712 `primaryType` does not determine the type dependency closure, preventing valid multi-type payloads from being signed.
- CR-005: Slippage quantization can loosen the user's minimum-receive threshold.
- CR-015: An older CLI does not reject future persistence schemas and can write them back using outdated semantics.
- CR-020: KDF parameters in imported V3 keystores have no computational resource limits.
- CR-024: Java fullnode-only configuration misclassifies a known network as `CUSTOM`.
- CR-026: After the CAIP-2 rename, the official EVM token validation test silently passes with an empty selection.

P3:

- CR-012: ADRs referenced by production code are not included, and their target directory is ignored.
- CR-013: Entropy-source failures are classified as `internal_error`, leaving `entropy_failure` unreachable.

ts/src/adapters/outbound/persistence/migration.ts
gummy789j: Confirmed and fixed in 6eec0a95.

`commit()` now re-reads each file under the lock and migrates what is on disk, rather than the snapshot `plan()` captured before the consent/password prompts. A version comparison is enough to detect interference: the gate is absolute, so another process cannot touch a stale file without migrating it first, which moves the version. Already migrated → skip; moved somewhere unexpected → refuse, leaving the file untouched. The pre-migration backup is taken from the fresh read too, so it can no longer be overwritten with the stale copy.

Reproduced with two real processes before fixing: A plans, B migrates and creates an account, A commits — B's account was gone from wallets.json and its key blob orphaned. After the fix it survives.

ts/src/application/use-cases/evm/transaction-service.ts
gummy789j: This one was already fixed before the review landed — cd39c9c7 ("fix(evm): stop tx status/info reporting a broken node as not_found") is an ancestor of the commit under review, so the behaviour described here is no longer present.

On the current code: line 430 is `gateway.getTransactionByHash(hash)` inside `Promise.all` with no catch, so transport and RPC failures propagate. The only `.catch(() => undefined)` is on the third, best-effort `getBlockNumber()` call, which contributes the `confirmations` field and cannot change `state`. One layer down, `evm.ts:406` distinguishes an explicit `result: null` (returns null) from a response missing `result` entirely (throws `invalid_node_response`), and `#call` turns any `body.error` into a `ChainError`.

No change made.

ts/src/adapters/outbound/chain/evm/signing-strategy.ts
gummy789j: Confirmed and fixed in cb9efd4d — and the scope was wider than this line: the same pattern was in all four signing paths (EVM software, EVM Ledger, TRON software, TRON Ledger).

Rather than fixing each adapter, `normalizeTypedData` now narrows `types` to the closure the declared `primaryType` reaches, so every path gets a single-root map. That also removes the Ledger paths' latent disagreement between the root `hash()` infers and the `primaryType` handed to `hashStruct()`.

Verified end to end: a payload with an unrelated extra type now signs, and to the same digest as the same payload without it — narrowing cannot change an existing signature, since every type a single-root payload declares is reachable from that root.

One case left as-is: multiple roots with NO `primaryType` still fails at signing. There is nothing to compute a closure from there, and changing it would move the failure from exit 1 to exit 2.

ts/src/adapters/inbound/cli/commands/exchange.ts
gummy789j: Confirmed and fixed in dcb227c9. Your worked example reproduces exactly: predicted 1,000,000 with 1.006% gave 989,900 against an exact 989,940.

Two things to add. The error runs in BOTH directions, because `Math.round` goes to nearest, not down: 1.004% produced 990,000 instead of 989,960, refusing a trade the user would have taken. And under 0.005% the floor snapped to the prediction itself — since that is a Bancor estimate, the trade then reverts on any last-unit disagreement, so asking for a very tight tolerance produced no tolerance at all. The code comment claimed "fractions finer than 0.01% round down"; the code did not implement that.

Fixed by moving to millionths of a percent (still integer arithmetic, exact for any precision a person types) and rounding the tolerance DOWN, so the floor can only come out stricter than requested. Every percentage basis points could already express is unchanged, including the spec's 1% example.

Took this over rejecting >2dp input: the value is computable, so refusing it would break working commands for no gain.

ts/.gitignore
gummy789j: Agreed on the problem, solved the other way round — 3bc95f78 removes the references instead of publishing the ADRs.

The complaint is right: comments pointed maintainers at documents they cannot retrieve. We decided not to version the ADR directory, so the citations had to go. 37 of them, in three spellings (`ADR-00XX`, `docs/adr/000X` paths, and one `ADR/deviations D6`). Where the citation was decoration the sentence already stood alone; where the ADR carried the reasoning, the reason is now inline — why `--path` is unimplemented, what makes the node-rejection allowlist narrow, which migration leans on the address re-encoding.

While sweeping we found `architecture-source-of-truth.md`, cited by `config.test.ts`, does not exist anywhere in the repository either. Same treatment.

`src/` and `test/` now contain no reference to an unversioned file.

ts/src/application/use-cases/wallet-service.ts
gummy789j: Accurate, but not changing it.

`entropy_failure` is indeed only reachable from `address-service.ts` today, and a `randomBytes` failure inside the keystore path would surface as `internal_error`. The prescription is what we are declining: routing entropy through an injectable boundary adds an abstraction to the domain for a condition Node effectively cannot produce — `randomBytes` failing means the OS CSPRNG is unavailable, at which point the wallet has larger problems than its error code.

If we revisit it, the minimal form is a try/catch at the keystore-encrypt call site mapping to `entropy_failure`, not a new port.

ts/src/domain/migration/index.ts
gummy789j: Correct as described, and deliberate — this is a known gap we decided to accept rather than a defect this PR introduced.

`planMigrations` filters on `storedVersion < currentVersion`, so a file written by a NEWER binary is left alone and startup continues; there is a test asserting exactly that (`the check is < , not !==`). The reasoning was that the alternative — failing closed on a future version — turns any accidental downgrade into a hard lockout, and we judged silently-left-alone the better default while the schema is young.

You are right that "left alone" is not the same as "never written back", which is the part that makes this uncomfortable. It is on the list to revisit; we did not want to change startup behaviour inside this release.

ts/src/adapters/inbound/cli/commands/exchange.ts
gummy789j: The behaviour is exactly as you describe, and it is the intended design — not changing it.

The TRON exchange protocol expresses "no floor" only as `expected = 1`; there is no null. Our decision was that the CLI should not invent a floor the user did not ask for, because any number we picked would be a price opinion we have no basis for, and a wrong one silently reverts trades. So the choice is surfaced instead of guessed:

- `exchange trade --help` states it in capitals: "WITH NEITHER FLAG THERE IS NO SLIPPAGE PROTECTION: the trade accepts any non-zero return at any price."
- the response carries a warning saying the same thing, which is what an agent reading json sees.
- `--min-received`, `--raw-min-received` and `--slippage` are all one flag away.

Your point about warnings not stopping an irreversible operation stands on its own terms. We weighed it against a default floor that would reject trades users intended to make, and chose the explicit-opt-in shape. Happy to revisit if you think the automated-caller case is strong enough to outweigh that.

ts/src/adapters/inbound/cli/commands/wallet.ts
gummy789j: Confirmed and fixed in 71bb21e6, with one correction to the mechanism.

The memory half is already covered: @noble/hashes refuses a non-power-of-two `n` outright and caps `128*r*(n+p+1)` at a 1 GiB `maxmem`. Measured, `n=2^30` is rejected in 0 ms — so scrypt cannot OOM us, and a tighter bound of ours would only risk refusing files that library accepts. scrypt's `n`/`r`/`p` are therefore left alone deliberately.

What genuinely had no ceiling was pbkdf2. `c` is unbounded (noble only checks `c >= 1`): measured, `c=10,000,000` takes 7.8 s and scales linearly, so `c=10^9` runs ~13 minutes. And `dklen` had only a lower bound while pbkdf2 accepts up to `(2^32-1)*32`: `dklen=2^34` did not finish at all. Both now bounded before the KDF runs, and there is a test asserting the refusal is immediate rather than the result of doing the work first.

One thing your comment prompted that we would otherwise have missed: noble's own refusals were escaping as bare `Error`s, i.e. surfacing as `internal_error`. A file naming impossible parameters was reading as a fault in the CLI. They are now `invalid_keystore` like every other malformed-file rejection.

Costs real wallets write (our own export, MEW's 262,144) are untouched.

ts/src/adapters/outbound/config/index.ts
gummy789j: Confirmed and fixed in 904fc403 — the mainnet outcome you describe is real.

The mechanism is the fold reaching BACKWARDS. It walks the whole book, builtin entries included, so an entry keyed on `tron:3448148188` gets followed and rewrites the builtin `nile -> tron:3448148188` to that key's own target. `--network nile` would then build, sign and broadcast on mainnet.

A canonical id is now a terminal target, never a hop. Regression test included; it fails without the guard.

One deviation from your prescription: we did not make canonical ids invalid as alias KEYS. `resolve()` consults ids before the book, so such an entry is unreachable and harmless once the fold stops following it — and there is an existing test asserting that a book entry shadowing a canonical id loses to the id rather than being rejected. Refusing it would have overturned that deliberate choice to fix a problem that lived entirely in the fold.

java/docs/reference/config.md
gummy789j: Confirmed and fixed in a9adaf15. Pre-existing defect rather than one this release introduced, but worth fixing where you found it.

`identifyNetwork` now compares only the endpoints the config actually supplied; an omitted one says nothing about which network this is and is not compared. Connection behaviour and the fallback are unchanged — and notably `ApiClient` already honours `isSoliditynodeEmpty` by switching to local-create, so identification was the one consumer ignoring a flag the rest of the code respects.

The consequences were worse than a wrong label: the GasFree commands refuse any network other than MAIN or NILE, so a user on Nile was told GasFree does not support their network, and the transaction history applies an extra endpoint filter once the network is CUSTOM.

Also corrected the config reference, which said the network is inferred by comparing both endpoints while listing `soliditynode.ip.list` as optional.

Unit tests cover the identification rule (they fail against the old comparison). A fullnode-only start against a live Nile node still wants one manual run — flagging that rather than claiming coverage we do not have.

ts/src/adapters/outbound/tokenbook/builtins.test.ts
gummy789j: Confirmed and fixed in f1a04133. Verified by running the file: it reported 4 passing tests, because the three `it.each(evmNetworks)` blocks over an empty selection produce no cases at all.

So none of it was running — not the EIP-55 checksums, not the explicit decimals, not the duplicate-contract rule. Exactly the checks that stand between a mistyped contract address and lost funds, which is what makes this worse than a stale filter.

Fixed the namespace to `eip155:` and added your suggestion: a case asserting the selection is non-empty and that at least one network carries tokens, so the next rename fails the suite instead of silently emptying it.

4 tests before, 13 after.

zerodevblock-cyber (approved):

PR #997, opened 2026-09-03
docs: expand agent multichain safety guidance

grayfoxd:


ts/docs/machine-interface.md
```diff
@@ -199,6 +199,13 @@ retry, exactly as the four-state model below requires.
 
 That index is the machine-readable catalog exposed by this build. Treat it as a discovery aid, not a closed enum: a few code paths choose among error-code strings dynamically, so a runtime envelope can still carry a code not present in `errorCodes`. The tables below are the frequently-hit subset, kept for reading. New codes may still be added within v1, and two strings (`invalid_value`, `aborted`) can appear under either exit code depending on where they are raised — so always tolerate an unknown code by falling back to its exit-code class.
 
+The catalog deliberately does not prescribe retry behavior. A transport `timeout` or `rpc_error`
+only says that the CLI did not observe a successful response. For a read-only call, retrying the same
+request is normally harmless. On a submit path, however, the node may have accepted the transaction
+before the response was lost. Treat that outcome as unknown: reconcile by `txId` when available, or
+by sender and nonce for EVM, before building and signing another transaction. Re-running a send does
+not retry the same transaction; it creates a second transaction and can duplicate a payment.
+
```
gummy789j, line 208: There is a section more clearly to explain the same thing

```
retry describes the error, not the command. timeout and rpc_error are same because
for most calls that is correct — the node never acted, so resending is free. But a command that
may have already broadcast a transaction (tx send and anything else on the submit path) can hit
timeout or rpc_error after the node accepted the transaction and before the response made it
back. In that case the outcome is unknown, not failed, and resending does not retry the original
request — it builds and signs a new transaction, which on TRON is a second, distinct transfer.
retry: "same" is correct for a timeout/rpc_error that happens while resolving a network id or
reading a balance; it is not a license to resend a broadcast blind. Reconcile with
[tx status](https://github.com/tronprotocol/wallet-cli/pull/997/changes#script-safety-never-mistake-submitted-for-confirmed) before deciding whether to
retry, exactly as the four-state model below requires.
```

PR #1004, opened 2026-09-09
Release v4.13.1

gummy789j:
## Notice

Non-mandatory upgrade

If you created HD accounts at index `1` or above on 4.13.0 or earlier, read
[Troubleshooting](#troubleshooting) before upgrading — those TRON addresses need one export step
that must happen while the old wallet is still intact.

### Change

1. **`derive` no longer demands a seed id.** It now selects the HD wallet the way every other account command does: `--account <accountId | label | address>` names any account of the wallet, and with neither flag the active account decides. `--seed-id wlt_…` still works and takes precedence, so an existing script is unaffected, while `wallet-cli derive` on its own is now the common case. Selecting an account that belongs to a private-key, watch-only or Ledger wallet fails with `seed_not_found` at exit `2` and says which flag to reach for; `--index` names the address index in `m/44'/<coin>'/0'/0/<index>`, and re-running it on an index that already exists remains a no-op that re-activates the slot. ([#1003](https://github.com/tronprotocol/wallet-cli/pull/1003))

2. **`backup` asks which format you want.** In an interactive terminal, omitting `--keystore` now opens a two-line choice — *Native wallet backup* (the recovery phrase for the whole HD wallet) or *Web3 keystore* (the single private key for the selected network's family) — instead of silently taking the native path. The prompt appears only for a fully interactive invocation: passing `--keystore`, or piping the password in with `--password-stdin`, keeps the long-standing non-interactive default, so pipes and CI stay deterministic. A native backup of a seed wallet now also warns about any account this version's default mnemonic recovery would not recreate, naming each one with the exact `backup --keystore` command that preserves it. ([#1003](https://github.com/tronprotocol/wallet-cli/pull/1003))

3. **`derivationPath` is now only ever reported by a command that actually opened the seed.** `list`, `current`, `use` and `rename` need no master password, and without the seed they cannot tell an account derived under the old TRON template from one derived under the corrected one — so rather than print the current template as though it were an observed fact, they now return `derivationPath: null` for every account. **A consumer that read `derivationPath` from those four commands must switch to a command that unlocks the seed:** `derive` and `backup` resolve each cached address against its actual current or historical template before reporting, so their value is verified rather than assumed. `ledger`, `watch` and `privateKey` accounts return `null` as before. ([#1003](https://github.com/tronprotocol/wallet-cli/pull/1003))

### Change

1. **HD accounts on TRON were derived at a path no other TRON wallet scans.** `derive --index 1` produced its TRON key at `m/44'/195'/1'/0/0`, hanging the number at BIP44's *account* level, while TronLink, the Java wallet-cli and the wider TRON ecosystem all increment *address_index*. The address was real and the funds are safe, but restoring the same recovery phrase in any other wallet would not show it. TRON now derives at `m/44'/195'/0'/0/<index>`, the same shape EVM already used, so a derived account follows the default restore flow everywhere. **Index `0` is identical under both templates and is unaffected**; only index `1` and above were reachable at the old path. Accounts already sitting there are detected rather than overwritten: the stored address is resolved against both templates before any key is used, so signing such an account fails with the new `legacy_derivation` code (exit `1`) pointing at the recovery guide instead of signing with a different key; deriving a further account into that wallet is refused for the same reason, because one wallet must never mix templates; the 4.13.0 → 4.13.1 wallet-file migration keeps a cached TRON address that the old template explains, rather than re-deriving over it; and `backup --keystore` deliberately still exports the old path's key, since that is the one command whose whole purpose is to let the key leave. A stored address that *no* template explains is a different failure with a different fix — `wallets.json` and the encrypted vault disagree — and now raises `derivation_mismatch` instead of being papered over. Both codes are published in `--json-schema`, and the complete ordered recovery procedure ships with the release. ([#1003](https://github.com/tronprotocol/wallet-cli/pull/1003))

2. **`import ledger --index` did not mean what the device shows.** The index was fed through the software derivation template, so on EVM it registered `m/44'/60'/0'/0/<index>` — an account Ledger Live does not display, leaving a user who picked "account 1" on their device with a different address than the one in front of them. Hardware paths are now built from Ledger Live's own template, `m/44'/<coin>'/<index>'/0/0`, for both families; TRON's path is unchanged, since that already was the template it used. `--path` remains the way to register any other scheme explicitly, `--address` scans Ledger Live accounts to find one, and the interactive account picker now prints each candidate's full path next to its address and says which template it is listing — so an account that lives on another scheme can no longer be silently mistaken for absent. ([#1003](https://github.com/tronprotocol/wallet-cli/pull/1003))

## Troubleshooting

**`legacy_derivation` — recovering TRON addresses derived before this release.**

If a TRON account was derived at index `1` or above by 4.13.0 or earlier, it lives at
`m/44'/195'/<index>'/0/0`, a path this version no longer produces. Signing it, or deriving another
account into the same wallet, now fails with `legacy_derivation` (exit `1`). The key is not lost —
the recovery phrase still derives it at that path — but this version's default import and derive
flow will not recreate the address, so it must be preserved before the wallet is rebuilt:

```console
error [legacy_derivation]: account "main-1" was derived at m/44'/195'/1'/0/0, a TRON path this
version no longer produces, so it cannot be signed here. Follow the complete recovery procedure
before deleting anything:
  https://github.com/tronprotocol/wallet-cli/blob/wallet-cli-4.13.1/ts/docs/troubleshooting/legacy-derivation-recovery.md
```

The full procedure — save the account list and recovery phrase, export each affected TRON account
with `backup --keystore` and re-import it as a standalone account, then delete and re-import the
mnemonic wallet and re-derive every index — is documented in
[**Recover addresses after `legacy_derivation`**](https://github.com/tronprotocol/wallet-cli/blob/wallet-cli-4.13.1/ts/docs/troubleshooting/legacy-derivation-recovery.md).
Ethereum derivation did not change, so every EVM address is restored unchanged by the re-import;
only TRON at index `1` and above needs the export step. **Do not delete the mnemonic wallet until
every TRON address you need appears as a standalone private-key account.** The whole procedure
reorganizes local keys only — it moves no funds on chain.

`derivation_mismatch` is a different failure and has no such rescue: it means the stored address
matches no derivation path of its seed, which an edited `wallets.json` or a wallet pointing at the
wrong vault would cause.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -108,9 +108,6 @@ cd java
 ./qa-repl/cli-boundary.sh build/libs/wallet-cli.jar    # entry point takes only --version/--help
 ```
 
-See `java/qa-repl/README.md`. The older `qa/` harness only ever drove the standard
-CLI, which was removed in v4.13.0.
-
 ## Architecture
 
 This is a **TRON blockchain CLI wallet** built on the [Trident SDK](https://github.com/tronprotocol/trident). It communicates with TRON nodes via gRPC.
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -19,7 +19,7 @@ plugins {
 }
 
 group 'Tron'
-version '4.13.0'
+version '4.13.1'
 
 apply plugin: 'java'
 apply plugin: 'com.google.protobuf'
--- a/java/src/main/java/org/tron/common/utils/Utils.java
+++ b/java/src/main/java/org/tron/common/utils/Utils.java
@@ -136,7 +136,7 @@ public class Utils {
 
   public static final int MIN_LENGTH = 2;
   public static final int MAX_LENGTH = 14;
-  public static final String VERSION = " v4.13.0";
+  public static final String VERSION = " v4.13.1";
   public static final String TRANSFER_METHOD_ID = "a9059cbb";
 
   private static SecureRandom random = new SecureRandom();
--- a/ts/docs/commands/backup.md
+++ b/ts/docs/commands/backup.md
@@ -18,7 +18,7 @@ Two formats:
 - **Native** (default) — the wallet's own backup JSON. A seed account exports its recovery phrase, so the whole seed moves with it.
 - **`--keystore`** — a standard Web3 keystore JSON, importable by TronLink and others, encrypted with **your master password**. A keystore holds a **single private key**: an HD account exports only its current derived key, and that key arrives elsewhere as a standalone account with nothing derivable from it. Use the native format to move a seed.
 
-The native export warns when the wallet contains an affected account whose TRON address uses the historical path `m/44'/195'/<n>'/0/0`. A normal mnemonic import in this version follows the current template and does not recreate that address automatically, so export each account named by the warning with `--keystore` before deleting it. See [Accounts & HD](../concepts/accounts-and-hd.md).
+The native export may warn that some stored accounts need a separate `--keystore` export. Follow that warning before deleting anything; see the [4.13.1 release notes](https://github.com/tronprotocol/wallet-cli/releases/tag/wallet-cli-4.13.1) for migration details.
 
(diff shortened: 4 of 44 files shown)


matrix-agent116 (approved):

zerodevblock-cyber (approved):

mollyvong:
Because of all of your working make a lots of people sadness, sometimes
they’re homeless, they’re are no home and sometimes people they’re are
steal everything from me . Do you think I am a machine?

Monny

